PR #11, opened 2019-12-04
Add "--enable-media-stream" command-line flag used to validate local …

thierryg:
…ressources such as a webcam.

This is quite usefull with webrtc app for instance such as http://apprtc.appspot.com/.

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -218,6 +218,7 @@ class App : public CefApp
                                              CefRefPtr<CefCommandLine> command_line) override
   {
     command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
+    command_line->AppendSwitch("enable-media-stream");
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
Seems OK to me. We really should have a property to allow specifying extra command line parameters, but this seems like a sane default. Thanks!

thierryg:
Indeed, it could be a default parameter to cef without any other flag for
cefsrc.

Regards

Thierry GAYET
Mobile : +33.663.849.589
Skype : tux35220

MathieuDuponchelle:
That's another question, but not what I meant here :)

thierryg:
yes, on having a property to pass command line parameters to cef it's an
interesting idea.

Regards

Thierry GAYET
Mobile : +33.663.849.589
Skype : tux35220

MathieuDuponchelle:
just needs doing really :D

PR #18, opened 2021-03-02
Use F32LE format for audio

reinismu:
From what I know big-endian is rare format. When I was writing custom recorder for cef my output was in F32LE

### DIFF
--- a/gstcefdemux.cc
+++ b/gstcefdemux.cc
@@ -4,7 +4,7 @@
 #include "gstcefaudiometa.h"
 
 #define CEF_VIDEO_CAPS "video/x-raw, format=BGRA, width=[1, 2147483647], height=[1, 2147483647], framerate=[1/1, 60/1]"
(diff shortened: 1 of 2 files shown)


MathieuDuponchelle:
we can't just pick the endianness though, we simply have to mirror whatever CEF is giving us, and I'm decently sure I had picked BE for a reason, though the audio capture API has changed since then

reinismu:
I currently run my version compiled using `F32LE`. I'm surprised that it works both ways tho

MathieuDuponchelle:
yes, I also tried and was surprised, I don't have a good explanation why either works rn :)

MathieuDuponchelle:
Closing for now, if I ever find time to understand what exactly is at play here I will comment :)

MathieuDuponchelle:
@reinismu I made sense of it, there was an inconsistency in the code, we ended up pushing F32LE caps anyway, that's why the two other places where F32BE was used didn't make a difference, apart from also making no sense :D

PR #19, opened 2021-03-02
Add enable-begin-frame-scheduling flag

reinismu:
This seems to help when running headless https://bitbucket.org/chromiumembedded/cef/issues/1368

Tested and it worked fine with it

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -307,6 +307,7 @@ class App : public CefApp
     command_line->AppendSwitch("disable-gpu");
     command_line->AppendSwitch("disable-dev-shm-usage"); /* https://github.com/GoogleChrome/puppeteer/issues/1834 */
     command_line->AppendSwitch("disable-gpu-compositing");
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
Interesting, the issue is a bit involved, but it looks like there's a correlation between usage of this flag and CPU usage, do you think a property would make sense?

reinismu:
I don't think it would give any harm. It is used in CasparCG as well https://github.com/CasparCG/server/blob/7ac31ee94f810a41c931e2000f11c874b7da7930/src/modules/html/html.cpp#L189

MathieuDuponchelle:
tnx, let me test locally :)

MathieuDuponchelle:
OK, fwiw on my machine I don't see a clear difference when using https://www.testufo.com/ as the source, but it also has no negative impact on the CPU

MathieuDuponchelle:
ooc, does this improve something on your end?

reinismu:
No clear difference. Could be it doesn't give so much performance. Tho it is mentioned in many places https://github.com/cztomczak/cefpython/issues/240
But ye up to you to put it in or not

MathieuDuponchelle:
well I took it, if everything breaks it will be your fault of course

PR #21, opened 2021-03-02
Add pr build test github action

reinismu:
Some nice to have thingy :) 

Later if there is a need could test that we can record a video as well

Tested here https://github.com/reinismu/gstcefsrc/pull/1

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -1,0 +1,39 @@
+name: gstcefsrc pull request test
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    name: Run gstcefsrc tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
nice, keep em coming :)

reinismu:
It seems that github actions are disables in this repo. You can enable them in settings

MathieuDuponchelle:
hm, am I looking in the wrong place? 
Screenshot from 2021-03-03 01-07-53

reinismu:
Thought it would run on this PR as well. I guess will have to check next PRs

reinismu:
Not sure if it needs more :thinking: Would be nice to have it in rust tho :) 
Now compiling cef with all codecs so could stream websites with video playing in background uhh takes so long

MathieuDuponchelle:
It would be nice yes, I have no idea if there exist complete enough bindings for that purpose, and I'm definitely not looking to implement any myself :)

PR #22, opened 2021-03-04
Add a gpu support property

reinismu:
With GPU 60 fps
image

Without GPU only 6-7 fps

I'm still very surprised that this works. Was under impression that GPU in osr mode works only under windows

Tested with this site https://threejs.org/examples/#webgl_points_dynamic

p.s. first c code I write in years

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -22,6 +22,7 @@ GST_DEBUG_CATEGORY_STATIC (cef_src_debug);
 #define DEFAULT_FPS_N 30
 #define DEFAULT_FPS_D 1
 #define DEFAULT_URL "https://www.google.com"
+#define DEFAULT_GPU FALSE
 
 static gboolean cef_inited = FALSE;
 static gboolean init_result = FALSE;
@@ -32,6 +33,7 @@ enum
 {
   PROP_0,
   PROP_URL,
+  PROP_GPU,
 };
 
 #define gst_cef_src_parent_class parent_class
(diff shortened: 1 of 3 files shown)


reinismu:
Github action worked as well :sunglasses:

gstcefsrc.cc
```diff
@@ -680,6 +696,11 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
           "The URL to display",
           DEFAULT_URL, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | G_PARAM_CONSTRUCT)));
 
+  g_object_class_install_property (gobject_class, PROP_GPU,
+    g_param_spec_boolean ("gpu", "gpu",
+          "Enable GPU",
```
MathieuDuponchelle, line 701: Please expand a bit on the description :)

gstcefsrc.cc
```diff
@@ -680,6 +696,11 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
...
+          FALSE, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS)));
```
MathieuDuponchelle, line 702: Not obvious, but this property will only be used when the element undergoes the READY -> PAUSED state change, so you want to indicate that it is `GST_PARAM_MUTABLE_READY` :)

MathieuDuponchelle, line 702: by convention, we usually stick all the default values at the top of the file as defines, see here `DEFAULT_URL`, can you do that? :)

MathieuDuponchelle:
Looks good to me otherwise, well done :) I'll give this a spin on my machine, I'm also surprised that it is possible on linux but that's good news :)

woot re action :)

gstcefsrc.cc
reinismu: Interesting... gstreamer change properties after it has been started?

gstcefsrc.cc
reinismu: I added a bit more information, but if you want it to tell something else, let me know ;)

gstcefsrc.cc
reinismu: Sure

MathieuDuponchelle: Well, the property mechanism can be used by whoever is using this element when the element is in any state through the API, when using an element with gst-launch the property will be set when the element is still in the NULL state, but gst-launch is not the sole user of gstreamer's API :) By using `GST_PARAM_MUTABLE_READY` you signal to whoever might call `g_object_set(prop, provalue)` on the element that their changes will not be used until the element cycles back to READY , if it is in a higher state :)

MathieuDuponchelle:
tnx, will give it a test later :)

reinismu:
> tnx, will give it a test later :)

Any luck on checking it out? :)

MathieuDuponchelle:
Hey @reinismu sorry this slipped my mind, and I was in vacation last week :)

Unfortunately I can't test this locally, I have a nvidia card and see nasty errors in my terminal:

```
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
```

The website loads fine but the animation isn't displayed.

In any case, this is property-gated and the code is fine, so given you see gains on your end it needs merging, thanks :)

PR #23, opened 2021-04-05
Allow loading a new url in the PLAYING state

rationalsa:
I'm not 100% sure when `gst_cef_src_init()` runs, is `state_lock` guaranteed to be initialized by the time I try to use it?


MathieuDuponchelle:
> I'm not 100% sure when `gst_cef_src_init()` runs, is `state_lock` guaranteed to be initialized by the time I try to use it?

An element is a GObject, so `init()` is called at instantiation, not a problem here. The patch makes sense to me and I can't imagine a deadlock at a glance, merging thanks :)

PR #26, opened 2021-06-09
Add params to set chromium-debug-port

reinismu:
image

Enables remote debugging for chromium

### DIFF
--- a/gstcefaudiometa.cc
+++ b/gstcefaudiometa.cc
@@ -37,7 +37,7 @@ gst_buffer_add_cef_audio_meta (GstBuffer * buffer, GstBufferList *buffers)
 GType
 gst_cef_audio_meta_api_get_type (void)
 {
-  static volatile GType type;
+  static GType type;
   static const gchar *tags[] = { NULL };
 
   if (g_once_init_enter (&type)) {
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -23,6 +23,7 @@ GST_DEBUG_CATEGORY_STATIC (cef_src_debug);
 #define DEFAULT_FPS_D 1
 #define DEFAULT_URL "https://www.google.com"
 #define DEFAULT_GPU FALSE
+#define DEFAULT_CHROMIUM_DEBUG_PORT -1
 
 static gboolean cef_inited = FALSE;
 static gboolean init_result = FALSE;
@@ -34,6 +35,7 @@ enum
   PROP_0,
   PROP_URL,
   PROP_GPU,
+  PROP_CHROMIUM_DEBUG_PORT,
 };
 
 #define gst_cef_src_parent_class parent_class
@@ -315,6 +317,10 @@ class App : public CefApp
       // Optimize for no gpu usage
       command_line->AppendSwitch("disable-gpu");
       command_line->AppendSwitch("disable-gpu-compositing");
+    }
+
+    if (src->chromium_debug_port >= 0) {
+      command_line->AppendSwitchWithValue("remote-debugging-port", g_strdup_printf ("%i", src->chromium_debug_port));
     }
   }
 
@@ -627,6 +633,11 @@ gst_cef_src_set_property (GObject * object, guint prop_id, const GValue * value,
       src->gpu = g_value_get_boolean (value);
       break;
     }
+    case PROP_CHROMIUM_DEBUG_PORT:
+    {
+      src->chromium_debug_port = g_value_get_int (value);
+      break;
+    }
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
       break;
@@ -645,6 +656,9 @@ gst_cef_src_get_property (GObject * object, guint prop_id, GValue * value,
       break;
     case PROP_GPU:
       g_value_set_boolean (value, src->gpu);
+      break;
+    case PROP_CHROMIUM_DEBUG_PORT:
+      g_value_set_int (value, src->chromium_debug_port);
       break;
     default:
       G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
@@ -679,6 +693,7 @@ gst_cef_src_init (GstCefSrc * src)
   src->audio_buffers = NULL;
   src->audio_events = NULL;
   src->started = FALSE;
+  src->chromium_debug_port = DEFAULT_CHROMIUM_DEBUG_PORT;
 
   gst_base_src_set_format (base_src, GST_FORMAT_TIME);
   gst_base_src_set_live (base_src, TRUE);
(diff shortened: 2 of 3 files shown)


gstcefaudiometa.cc
```diff
@@ -37,7 +37,7 @@ gst_buffer_add_cef_audio_meta (GstBuffer * buffer, GstBufferList *buffers)
 GType
 gst_cef_audio_meta_api_get_type (void)
 {
-  static volatile GType type;
+  static GType type;
```
reinismu, line 40:
```
In file included from /usr/include/glib-2.0/glib/gthread.h:32,
                 from /usr/include/glib-2.0/glib/gasyncqueue.h:32,
                 from /usr/include/glib-2.0/glib.h:32,
                 from /usr/include/gstreamer-1.0/gst/gst.h:27,
                 from /home/reinismu/projects/gstcefsrc/gstcefaudiometa.h:4,
                 from /home/reinismu/projects/gstcefsrc/gstcefaudiometa.cc:1:
/home/reinismu/projects/gstcefsrc/gstcefaudiometa.cc: In function ‘GType gst_cef_audio_meta_api_get_type()’:
/usr/include/glib-2.0/glib/gatomic.h:117:19: error: argument 2 of ‘__atomic_load’ must not be a pointer to a ‘volatile’ type
  117 |     __atomic_load (gapg_temp_atomic, &gapg_temp_newval, __ATOMIC_SEQ_CST); \
      |     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I got this error when compiling

Did a bit of googling and found https://github.com/vmware/open-vm-tools/commit/b2c8baeaa8ac365e1445f941cf1b80999ed89a9d

gstcefsrc.cc
```diff
@@ -709,6 +723,11 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
           "Enable GPU usage in chromium (Improves performance if you have GPU)",
           DEFAULT_GPU, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
 
+  g_object_class_install_property (gobject_class, PROP_CHROMIUM_DEBUG_PORT,
+    g_param_spec_int ("chromium-debug-port", "chromium-debug-port",
+          "Set chromium debug port (0 = disabled)", 0, G_MAXUINT16,
+          DEFAULT_CHROMIUM_DEBUG_PORT, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
```
reinismu, line 729: For some reason default value always shows up as `0` when I do gst-inspect. I wanted to go with `-1`, but it didn't work :( 
Checked that someone else had had a similar experience, but didn't find a reason why (http://gstreamer-devel.966125.n4.nabble.com/why-I-can-t-use-g-param-spec-int-set-default-value-td4660762.html)
So went with 0

ystreet: You have to set the field to the default value in the `_init()` function.

gstcefaudiometa.cc
ystreet: https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/commit/98249a57dbd57dffbda12bfa9cb841fa41b28bbd

gstcefsrc.cc
reinismu: Hmm not sure I understand. Need a bit of spoonfeeding :D . Could you point to the place?
From what I see `url` prop is done in similar way and it works

MathieuDuponchelle: https://developer.gnome.org/gobject/stable/howto-gobject-construction.html:

> /* initialize all public and private members to reasonable default values.
>  * They are all automatically initialized to 0 to begin with. */

This works with URL because 0 is NULL :)

Besides that, gst-inspect by default will get_property to determine the default, there is an optional flag you could set for it to show the "actual" default, but this isn't relevant here.

Shouldn't we only allow >= 1024 though? In that case 0 would be a fair default anyway

gstcefaudiometa.cc
MathieuDuponchelle: damn it, the boilerplate is out of date :D @reinismu if you care to provide a commit to fix this it'll be welcome :)

gstcefsrc.cc
reinismu: Could change the range, but then what would be the disabled port value? :thinking:

gstcefaudiometa.cc
reinismu: Where is this boilerplate? :)

gstcefsrc.cc
MathieuDuponchelle: hah, good point good point, `G_MAXUINT16 + 1` :P

MathieuDuponchelle: but yeah no, -1 -> G_MAXUINT16, -1 disabled sounds fine

gstcefaudiometa.cc
MathieuDuponchelle: well, that volatile bit is boilerplate, copy pasted from somewhere else, just need to be removed really :)

gstcefsrc.cc
reinismu: But now how can I put the default value as -1? :D If I just change `DEFAULT_CHROMIUM_DEBUG_PORT` to -1 it doesnt work :( 
I checked https://developer.gnome.org/gobject/stable/howto-gobject-construction.html and it just puts `2` as default value there. Or am I missing something? :D 
If I put value in `gst_cef_src_init` then prop is not shoving at all with gst-inspect

MathieuDuponchelle: you need to set the minimum as -1 when registering the property

MathieuDuponchelle: it's really not that complicated, maybe go read the doc for install_property, and remember that gst-inspect-1.0 will *not* show the default you provide there, but instead g_object_get() the value after istantiating the element (for reasons, which you shouldn't worry about :P)

reinismu: Ok hmm I added `src->chromium_debug_port = DEFAULT_CHROMIUM_DEBUG_PORT;` as well and now it seems to work. 
So `This works with URL because 0 is NULL :)` how come? I'm so confused about this design decision. Why my `-1` doesn't get picked up but a string does? :thinking:

reinismu: And seems like chromium is ignoring 0 port value by default.

ystreet: Because the default initialization of an object (`g_object_new` effectively) sets everything to 0/NULL.  If you need something else in the object, you need to explicitly set that value inside `_init()`.

MathieuDuponchelle: as @ystreet said, @reinismu I feel for you if you're not used to writing (gobject) C, but if that comforts you there's no magic involved :)

Changes look OK now, can you fix up your commits, rebase on master and push again? I will merge afterwards :)

reinismu: To me, magic feels that how the default URL is set without being inside `_init()` part :D

MathieuDuponchelle:
thanks @reinismu !

PR #27, opened 2021-06-09
cefaudiometa: Remove incorrect use of volatile

ocrete:
Newer GCC complains with:
```
error: argument 2 of ‘__atomic_load’ must not be a pointer to a ‘volatile’ type
```


MathieuDuponchelle:
tnx :)

Issue #28, opened 2021-06-10
cefsrc and no GPU


PrzemekGajos:
Hi All,

I'm trying to use cefsrc without drawing to a screen **and without a GPU in my hardware**. Does anyone know if GPU is a requirement for cefsrc or can it be run in a headless mode?

For an example, (note that I'm using a virtual environment with Xvfb), and a simple test pipeline produces:
Xvfb :99 &
export DISPLAY=:99

GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 cefsrc num-buffers=1 url=file:///home/root/urls/text.html ! video/x-raw,format=BGRA,width=1920,height=1080,framerate=60/1 ! filesink location=/home/root/buffer.raw
Setting pipeline to PAUSED ...
Pipeline is live and does not need PREROLL ...
Setting pipeline to PLAYING ...
New clock: GstSystemClock
[1223/133936.433163:FATAL:gpu_data_manager_impl_private.cc(407)] GPU process isn't usable. Goodbye.
Trace/breakpoint trap

Thanks,
Przemek

MathieuDuponchelle:
@PrzemekGajos I don't see why this would happen no, if you don't set the "gpu" property at least we pass "disable-gpu" and "disable-gpu-compositing" to CEF initialization.

Googling for that error, I find https://bugs.launchpad.net/ubuntu/+source/chromium-browser/+bug/1877173 , which in turn links to other reports with several suggested workarounds, can you test those and report your findings?

PrzemekGajos:
@MathieuDuponchelle I'm testing those workarounds and I'll inform on the outcome.

Some of them suggest using '--no-sandbox'. 
Is there a way to pass command-line arguments to cefsrc at run time or would it be embedding them in the source code, e.g. where 'disable-gpu' is set?

MathieuDuponchelle:
@PrzemekGajos hardcode them and recompile, then we can expose properties if needed :)

MathieuDuponchelle:
@PrzemekGajos I'm going to close this for lack of feedback, please reopen if you have more info ofc :)

ghostnumber7:
[...]

PR #30, opened 2021-08-12
implement make install

ghostnumber7:
Related to talks on #28
I ended up implementing `make install` functionality instead of some custom variable. For example to install in `/usr/lib/x86_64-linux-gnu/gstreamer-1.0/gstcef/` you could run as follows
```shell
cmake \
      -DCMAKE_INSTALL_PREFIX=/usr/lib/x86_64-linux-gnu/gstreamer-1.0/gstcef/ \
      -DCMAKE_BUILD_TYPE=Release \
      -G"Unix Makefiles" .. && make && sudo make install
```
If that is GStreamer plugin path (can be a subdirectory as gstreamer recurses by default) then there is no need to set `GST_PLUGIN_PATH` variable, if not, you can simply set the variable to the install path as in `GST_PLUGIN_PATH=/usr/lib/x86_64-linux-gnu/gstreamer-1.0/gstcef/:$GST_PLUGIN_PATH` instead of using the `build/Release` path.

If `CMAKE_INSTALL_PREFIX` is not defined when building it will behave just like now (can't be installed and won't work if moved)

Tested on Debian (bullseye) and Ubuntu (focal) only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ project(gstcef)
 
 set_property(GLOBAL PROPERTY OS_FOLDERS ON)
 
+if(POLICY CMP0074)
+    #policy for <PackageName>_ROOT variables
+    cmake_policy(SET CMP0074 NEW)
+endif()
+
 set(CEF_VERSION "88.2.9+g5c8711a+chromium-88.0.4324.182")
 set(CEF_ESCAPED_VERSION "88.2.9%2Bg5c8711a%2Bchromium-88.0.4324.182")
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,6 +8,16 @@ cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release ..
...
+## Build + install
+You can build for install using any path. If you install inside GStreamer plugin path there is no need to set `GST_PLUGIN_PATH` variable.
+
+```
+mkdir build && cd build
+cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/lib/x86_64-linux-gnu/gstreamer-1.0/cefsrc/ ..
+make
+make install
+```
+
```
reinismu, line 20: Just for someone who hasn't experienced this issue I would add a `Troubleshooting` section that goes something like this
## Troubleshooting

#### Error: `GPU process isn't usable. Goodbye.`

The subprocess bin path is configured with the absolute path when building. Then that's called at runtime to spawn the subprocess or alike, so no matter where the other files are it will spawn that bin from build path only. 

To solve this use `-DCMAKE_INSTALL_PREFIX` as stated above.

More #28

reinismu (approved):
Awesome! :) 

I wonder if there is a way to know `libcef.so` location in runtime. Then could just use folders relative to it. 

But this works :)

ghostnumber7:
I guess there must be a way to either 1) know the location of `libgstcef.so` and infer from there or 2) install `libcef.so` in a proper lib path. Both approaches should work. I'll add the entry in troubleshooting tomorrow

CMakeLists.txt
```diff
@@ -73,12 +78,39 @@ SET_LIBRARY_TARGET_PROPERTIES("gstcef")
 add_dependencies("gstcef" libcef_dll_wrapper)
 target_link_libraries("gstcef" libcef_lib libcef_dll_wrapper ${CEF_STANDARD_LIBS} ${GST_LIBRARIES})
 target_include_directories("gstcef" PUBLIC ${GST_INCLUDE_DIRS})
-target_compile_definitions("gstcef" PUBLIC "-DCEF_SUBPROCESS_PATH=\"$<TARGET_FILE:gstcefsubprocess>\"")
-target_compile_definitions("gstcef" PUBLIC "-DCEF_LOCALES_DIR=\"${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/locales\"")
+
+if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  set(CEF_SUBPROCESS_BASE_PATH "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}")
+else()
+  set(CEF_SUBPROCESS_BASE_PATH "${CMAKE_INSTALL_PREFIX}")
```
MathieuDuponchelle, line 85: I think this means cefsrc won't be usable "uninstalled" when install prefix doesn't it?

ghostnumber7: yes. The thing with absolute path is that it will work either installed or uninstalled but never both. The only thing I can think of that could allow for both ways of usage would be to get the path somehow in the code instead of using a compile definition. I don't know how to achieve that at all :(

ghostnumber7: GstPlugin struct has a basename prop that should be what we need, but I don't know how to access it. If we get that value it should work with both installed and uninstalled. I'll see if I can get it somehow (maybe you know how to and can spare me the research?)

MathieuDuponchelle: good idea, `g_path_get_basename (gst_plugin_get_filename (plugin))` ?

ghostnumber7: yep it can be achieved with `g_path_get_dirname(gst_plugin_get_filename(plugin)`. The only problem I have now is that I don't see how to get the plugin at all in `gstcefsrc.cc` file (where I need to construct the path), and I can't see any way to send the information from `gstcef.cc` (where I do have the plugin). I can get over that by iterating `gst_registry_get_plugin_list(gst_registry_get())` to get the plugin though. Do you know of a way to get the plugin or pass the path where it's needed, or think using the registry would be fine?

MathieuDuponchelle: I think you want `gst_plugin_feature_get_plugin(GST_PLUGIN_FEATURE (gst_element_factory_find ("cefsrc")));`

ghostnumber7: I arrived to `gst_registry_find_plugin(gst_registry_get(), "cef");` which seams to do the same thing in this case. I'll do a few more tests.

MathieuDuponchelle: yes, that should be equivalent

ghostnumber7: I'll close this PR and open a new one with the other aproach as it solves the issue much better (making the plugin fully relocatable), if that's ok with you :)

PR #36, opened 2021-11-28
Small demux cleanup

reinismu:
For fun and learning, I started to rewrite `cefsrc` to Rust (Will see how far I get). Noticed that some fields are not used and a typo :)

### DIFF
--- a/gstcefdemux.cc
+++ b/gstcefdemux.cc
@@ -9,7 +9,7 @@
 G_DEFINE_TYPE (GstCefDemux, gst_cef_demux, GST_TYPE_ELEMENT);
 
 static GstStaticPadTemplate gst_cef_demux_sink_template =
-GST_STATIC_PAD_TEMPLATE ("src",
+GST_STATIC_PAD_TEMPLATE ("sink",
     GST_PAD_SINK,
     GST_PAD_ALWAYS,
     GST_STATIC_CAPS (CEF_VIDEO_CAPS)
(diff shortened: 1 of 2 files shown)


MathieuDuponchelle:
Hah, thanks :) Re rust, that sounds fun, one of the reasons I went for C++ was that I could simply copy paste the build system bits from CEF samples, how do you plan to deal with that? Does CEF have rust bindings by the way?

MathieuDuponchelle:
By the way, I would recommend not going for the same src / demux design, for cefsrc it was pretty much an ad-hoc decision in order to not have to rewrite the `GstBaseSrc`-based cefsrc when adding audio support, but if you start from scratch I would recommend simply implementing a `GstElement` subclass :)

reinismu:
> Hah, thanks :) Re rust, that sounds fun, one of the reasons I went for C++ was that I could simply copy paste the build system bits from CEF samples, how do you plan to deal with that? Does CEF have rust bindings by the way?

I did solve the audio and video grabbing part before in Rust when I was still thinking of using FFMPEG. I think that shouldn't be hard :)

> By the way, I would recommend not going for the same src / demux design, for cefsrc it was pretty much an ad-hoc decision in order to not have to rewrite the `GstBaseSrc`-based cefsrc when adding audio support, but if you start from scratch I would recommend simply implementing a `GstElement` subclass :)

I had a feeling it could be done differently while looking at NdiSrcDemux, but currently, it might be too much for me to handle :D At least I know that this design works :D 
Hmm you know a similar base plugin I could look at to inspire from?

Issue #37, opened 2021-12-03
Cannot build the project from the 93.1.11+g9e254fa+chromium-93.0.4577.63 release


thierrygayet:
Hi,

I have tried to upgrade the latest stable release of the libcef in the CMakeLists.txt file and it builds up to release **92.0.27+g274abcf+chromium-92.0.4515.159**:

```
set(CEF_VERSION "93.1.11+g9e254fa+chromium-93.0.4577.63")
set(CEF_ESCAPED_VERSION "93.1.11%2Bg9e254fa%2Bchromium-93.0.4577.63")
```

From release **93.1.11+g9e254fa+chromium-93.0.4577.63**, the following error appears:

```
$ cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Release .. && make
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Downloading /home/tgayet/gstcefsrc/third_party/cef/cef_binary_93.1.11%2Bg9e254fa%2Bchromium-93.0.4577.63_linux64.tar.bz2.sha1...
-- Downloading /home/tgayet/gstcefsrc/third_party/cef/cef_binary_93.1.11%2Bg9e254fa%2Bchromium-93.0.4577.63_linux64.tar.bz2...
-- [download 0% complete]
-- [download 1% complete]
-- [download 2% complete]
-- [download 3% complete]
-- [download 4% complete]
-- [download 5% complete]
-- [download 6% complete]
-- [download 7% complete]
-- [download 8% complete]
-- [download 9% complete]
-- [download 10% complete]
-- [download 11% complete]
-- [download 12% complete]
-- [download 13% complete]
-- [download 14% complete]
-- [download 15% complete]
-- [download 16% complete]
-- [download 17% complete]
-- [download 18% complete]
-- [download 19% complete]
-- [download 20% complete]
-- [download 21% complete]
[...]

PR #42, opened 2022-02-25
cmake, cefsrc: Update to Chromium 98.0.4758.102 + Bind API migration

fraxinas:
Required a trivial fix and a smol API migration


MathieuDuponchelle:
Hehe, very nice thanks, that was the fix for https://github.com/centricular/gstcefsrc/issues/37 then :)

PR #47, opened 2022-08-08
cefbin: Change protocol handler to web+...://

philn:
The URIs this element receives are not compliant with RFC 3986 because they have
2 protocol schemes (web://http://...). So we can't use the GstURI parser in
cefbin, we pass the unparsed URI stripped of the web:// or cef:// prefix to
cefsrc.

V2:

    These URIs are RFC 3986 compliant, so allowing for clean removal of the web+
    prefix before passing the URI to cefsrc.

### DIFF
--- a/gstcefbin.cc
+++ b/gstcefbin.cc
@@ -13,7 +13,7 @@ gst_cef_bin_uri_get_type (GType type)
 static const gchar *const *
 gst_cef_bin_get_protocols (GType type)
 {
-  static const gchar *protocols[] = { "web", "cef", NULL };
+  static const gchar *protocols[] = { "web+http", "web+https", "web+file", NULL };
 
   return protocols;
 }
@@ -36,11 +36,20 @@ gst_cef_bin_set_uri (GstURIHandler * handler, const gchar * uristr,
   gboolean res = TRUE;
   gchar *location;
   GstCefBin *self = GST_CEF_BIN (handler);
+  const gchar *protocol;
+  GstUri *uri;
 
-  location = gst_uri_get_location (uristr);
+  protocol = gst_uri_get_protocol (uristr);
+  g_return_val_if_fail (g_str_has_prefix (protocol, "web+"), FALSE);
+
+  uri = gst_uri_from_string (uristr);
+  gst_uri_set_scheme (uri, protocol + 4);
+
+  location = gst_uri_to_string (uri);
 
   g_object_set (self->cefsrc, "url", location, NULL);
 
+  gst_uri_unref (uri);
   g_free (location);
 
   return res;
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ run the previous commands with `xvfb-run`:
 `xvfb-run --server-args="-screen 0 1920x1080x60" gst-launch-1.0 ...`
 
 In addition, a wrapper bin is exposed, wrapping cefsrc and cefdemux, and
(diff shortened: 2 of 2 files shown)


MathieuDuponchelle:
Yes, I noticed this divergence with wpesrc too, however there is an alternative solution here, which is to not expect the "inner" protocol, and pick it / prepend it ourselves instead. I suppose we could have two variants of the web protocol then, `web` and `webs`. I honestly don't know what's the solution we should prefer, but that one is at least 3986-compliant :)

philn:
But if we forge a new URI I still don't see much point in using the GstURI parser. Not sure either what's the best approach :)

MathieuDuponchelle:
> But if we forge a new URI I still don't see much point in using the GstURI parser. Not sure either what's the best approach :)

Consistency I suppose, but more importantly we do want to harmonize `wpesrc` and `cefsrc` :)

If you don't mind updating `wpesrc` I think I slightly prefer my solution (no double protocols in URI seems correct), if not I guess we'll have to ask someone else to flip a coin :D

philn:
> Yes, I noticed this divergence with wpesrc too, however there is an alternative solution here, which is to not expect the "inner" protocol, and pick it / prepend it ourselves instead. I suppose we could have two variants of the web protocol then, `web` and `webs`. I honestly don't know what's the solution we should prefer, but that one is at least 3986-compliant :)

What about the `cef` scheme? We drop iti?

philn:
@thiblahute wdyt?

philn:
I kinda liked `web://proto://` because then we can use either http or file URIs...

MathieuDuponchelle:
@philn the idea would be to have `web` and `webs` instead :)

philn:
> @philn the idea would be to have `web` and `webs` instead :)

yes, i understood that part :) but then what if we want to load local resources too? How do we know when to generate a file:// or a https:// ?

MathieuDuponchelle:
Oh sorry I didn't read your comment right, yes true that's a problem

thiblahute:
Isn't the common way to do it to have a `web+http://` protocol ? (I have the impression that I have seen that notation in several places).

philn:
Sounds like the way to go! Hopefully the GstURI parser will no choke on that.

MathieuDuponchelle:
`scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`

Perfect, thanks thib :)

gstcefbin.cc
```diff
@@ -13,7 +13,7 @@ gst_cef_bin_uri_get_type (GType type)
 static const gchar *const *
 gst_cef_bin_get_protocols (GType type)
 {
-  static const gchar *protocols[] = { "web", "cef", NULL };
```
MathieuDuponchelle, line 16: I would like to keep the `cef` option too, when users want more control over the web backend :)

philn: can't the play with GST_PLUGIN_FEATURE_RANK?

MathieuDuponchelle: ah yes, I remember now, when I exposed this I didn't know there was an env var for ranks, only discovered that recently :) nvm me then.

MathieuDuponchelle:
Thanks!

philn:
I'll try to submit the corresponding change for wpesrc soon. Thanks for the reviews!

MathieuDuponchelle:
please do yes :+1:

PR #48, opened 2022-08-11
cefsrc: Relay JS console messages to GST_DEBUG facilities

philn:


pldin601 (approved):

gstcefsrc.cc
```diff
@@ -228,17 +228,36 @@ class AudioHandler : public CefAudioHandler
...
+class DisplayHandler : public CefDisplayHandler {
+public:
+  DisplayHandler(GstCefSrc *element) : mElement(element) {}
+
+  ~DisplayHandler() = default;
+
+  virtual bool OnConsoleMessage(CefRefPtr<CefBrowser>, cef_log_severity_t level,
+                                const CefString &message,
+                                const CefString &source, int line) override {
+    GST_DEBUG_OBJECT (mElement, "%s:%d : %s", source.ToString().c_str(), line, message.ToString().c_str());
```
philn, line 240: Ah depending on the level notified we could use GST_INFO or other gst logging level.

philn:
CI failure looks un-related. Some missing pkgs in some Azure Ubuntu mirror it seems.

gstcefsrc.cc
```diff
@@ -228,17 +228,56 @@ class AudioHandler : public CefAudioHandler
...
+    case LOGSEVERITY_DEBUG:
+      gstLevel = GST_LEVEL_DEBUG;
+      break;
+    case LOGSEVERITY_WARNING:
+      gstLevel = GST_LEVEL_WARNING;
+      break;
+    case LOGSEVERITY_ERROR:
+    case LOGSEVERITY_FATAL:
+      gstLevel = GST_LEVEL_ERROR;
+      break;
```
reinismu, line 254: This is for `console.error` right? Wouldn't it spam gst console? Not sure if I would put them in the same category

philn: What do you suggest then?

reinismu: I think they all should go under `GST_LEVEL_DEBUG` or `GST_LEVEL_INFO`. Depends on the use case I guess. In my case when I run my pipeline using third party website I wouldn't want to see their website logs in my gst output.

Maybe `GST_LEVEL_INFO` with a prop to turn it on or just `GST_LEVEL_DEBUG`.

What is your use case to have this? The website works fine in chrome and `cefsrc` have weird behavior?

philn: Yeah. I wonder, how about using a dedicated gst log category, like `cefconsole`? So that filtering might be easier, like `cefsrc:6,cefconsole:4` for instance.

MathieuDuponchelle:
Is this related to https://github.com/centricular/gstcefsrc/pull/49 or is the topic similarity coincidental? :)

philn:
> Is this related to #49 or is the topic similarity coincidental? :)

IIUC #49 is about setting the Chromium logging level, and this PR is about getting access to the JS console logs, so totally orthogonal :)

gstcefsrc.cc
MathieuDuponchelle: I think the separate category makes this fine, let's revisit if this proves too spammy :)

MathieuDuponchelle:
Thanks!

PR #49, opened 2022-08-11
cefsrc: Enable definition of cef log severity

aiden-loroco:
Allows users to define the log_severity setting in cef to `verbose / debug | info | warning | error | fatal | disable[default]` (corresponding to `cef_log_severity_t` options [here](https://magpcss.org/ceforum/apidocs3/projects/(default)/cef_types.h.html)).

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -2,12 +2,6 @@
 #include <iostream>
 #include <sstream>
 
-#include <include/cef_app.h>
-#include <include/cef_client.h>
-#include <include/cef_render_handler.h>
-#include <include/cef_life_span_handler.h>
-#include <include/cef_load_handler.h>
-#include <include/wrapper/cef_helpers.h>
 #include <include/base/cef_bind.h>
 #include <include/base/cef_callback_helpers.h>
 #include <include/wrapper/cef_closure_task.h>
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -19,12 +19,37 @@ GST_DEBUG_CATEGORY_STATIC (cef_src_debug);
 #define DEFAULT_URL "https://www.google.com"
 #define DEFAULT_GPU FALSE
 #define DEFAULT_CHROMIUM_DEBUG_PORT -1
+#define DEFAULT_LOG_SEVERITY LOGSEVERITY_DISABLE
 #define DEFAULT_SANDBOX TRUE
 
 static gboolean cef_inited = FALSE;
 static gboolean init_result = FALSE;
 static GMutex init_lock;
 static GCond init_cond;
+
+#define GST_TYPE_CEF_LOG_SEVERITY_MODE \
+  (gst_cef_log_severity_mode_get_type ())
+
+
+static GType
+gst_cef_log_severity_mode_get_type (void)
+{
+  static GType type = 0;
+  static const GEnumValue values[] = {
+    {LOGSEVERITY_DEBUG, "debug / verbose cef log severity", "debug"},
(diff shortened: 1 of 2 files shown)


gstcefsrc.cc
```diff
@@ -2,12 +2,8 @@
 #include <iostream>
 #include <sstream>
 
-#include <include/cef_app.h>
```
MathieuDuponchelle, line 5: I can't tell if the modified includes have any impact, but they shouldn't be part of this commit in any case :)

gstcefsrc.cc
```diff
@@ -830,6 +844,12 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
           "(Example: --noexpose_wasm --expose-gc)",
           NULL, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
 
+  g_object_class_install_property (gobject_class, PROP_LOG_SEVERITY,
+    g_param_spec_string ("log-severity", "log-severity",
```
MathieuDuponchelle, line 848: You want to register an enum here, and install the property as `g_param_spec_enum` :)

MathieuDuponchelle:
Thanks, that seems useful, but I don't think the PR is quite right yet :)

sidsethupathi:
From #48 
> Is this related to #49 or is the topic similarity coincidental? :)

I think it's coincidental :)

@aiden-loroco and I were trying to get `cefsrc` running in a Docker container but the `gst-launch` command kept exiting without displaying anything useful (it just logged out `INFO cefsrc gstcefsrc.cc:423:run_cef: Initializing CEF` and then exited).

Once we turned `settings.log_severity = LOGSEVERITY_DEBUG`, we got something that helped us figure out we needed to set `sandbox=false` (`Running as root without --no-sandbox is not supported. See https://crbug.com/638180.`).

gstcefsrc.cc
aiden-loroco: These are already included in the header. I'll split out into another commit.

gstcefsrc.cc
aiden-loroco: Nice, that's much neater. I should've known there was gobject way!

gstcefsrc.h
```diff
@@ -1,6 +1,8 @@
 #ifndef __GST_CEF_SRC_H__
 #define __GST_CEF_SRC_H__
 
+#include <glib.h>
```
MathieuDuponchelle, line 4: That shouldn't be needed, `gst.h` will bring it in

gstcefsrc.cc
```diff
@@ -824,6 +858,12 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
           "(Example: --noexpose_wasm --expose-gc)",
           NULL, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
 
+  g_object_class_install_property (gobject_class, PROP_LOG_SEVERITY,
+      g_param_spec_enum ("log-severity", "log-severity",
+          "CEF log severity level ",
+          GST_TYPE_CEF_LOG_SEVERITY_MODE, DEFAULT_LOG_SEVERITY,
+          (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS)));
```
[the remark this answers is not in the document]
aiden-loroco: Thanks!

MathieuDuponchelle:
All good now let's go, thanks :muscle:

PR #51, opened 2022-08-22
cmake: Allow user-supplied CEF_VERSION override

philn:
if not provided with `-DCEF_VERSION=...` fallback to hardcoded version number.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,10 @@ if(POLICY CMP0074)
     cmake_policy(SET CMP0074 NEW)
 endif()
 
-set(CEF_VERSION "103.0.9+gd0bbcbb+chromium-103.0.5060.114")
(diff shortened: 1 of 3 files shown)


philn:
The context of this PR is to allow custom CEF builds, that could be done with https://github.com/philn/cef-build-box for instance.
The Spotify CEF builds (obviously) don't ship with H.264 codec support, so in some cases like this a custom build is required.

MathieuDuponchelle:
heh, that seems useful thanks. For the record I think you could already use a custom *local* build with  `CEF_ROOT`, but the ability to fetch a custom build from the net is obviously nice to have :)

philn:
Yeah, it was quite a pain in the butt :( +1 to mention both our cef build container projects in README :)

MathieuDuponchelle:
go for it :)

PR #53, opened 2022-09-14
cefsrc: Allow custom CEF subprocess wrappers

philn:
If an override is supplied with the GST_CEF_SUBPROCESS_PATH environment variable, we set CEF_SUBPROCESS_PATH to the real subprocess path and spawn the supplied override program instead. It is then free of configuring the subprocess for cgroups, for instance.


MathieuDuponchelle:
I see, that sounds useful, thanks :)

PR #57, opened 2022-11-02
cefsrc: Add cef_cache_location property

reinismu:
When recording same website multiple times it helps to have most resources cached :) 

Now looking at `CefSettings` I wish we could just pass json and fill it up. Could solve all similar PRs down the line.

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -63,6 +63,7 @@ enum
   PROP_SANDBOX,
   PROP_JS_FLAGS,
   PROP_LOG_SEVERITY,
+  PROP_CEF_CACHE_LOCATION,
 };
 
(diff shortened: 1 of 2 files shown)


MathieuDuponchelle:
Yes, mapping all fields as properties is a little bit obnoxious. The only alternative I can think of would be to register a new boxed type around `CefSettings`, and then registering serialization / deserialization functions with `gst_value_register` for command line usage. That's a bit complex though :)

MathieuDuponchelle:
I'm happy with the patch as it is anyway, merging, but @reinismu if you feel like playing around with the idea you can find the entry point for an example here: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-editing-services/ges/ges.c#L166

PR #60, opened 2022-11-25
cefsrc: Apply running time on audio buffers

philn:
Also add an audio meta to buffers, for good measure, and set discont flag when needed. This should help with A/V sync issues and audio cracks reported in #59.

### DIFF
--- a/gstcefdemux.cc
+++ b/gstcefdemux.cc
@@ -89,6 +89,58 @@ typedef struct
   GstFlowReturn combined;
 } AudioPushData;
 
+#if !GST_CHECK_VERSION(1, 18, 0)
+static GstClockTime
+gst_element_get_current_clock_time (GstElement * element)
+{
+  GstClock *clock = NULL;
+  GstClockTime ret;
+
+  g_return_val_if_fail (GST_IS_ELEMENT (element), GST_CLOCK_TIME_NONE);
+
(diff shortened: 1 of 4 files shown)


pldin601 (approved):

philn:
CI failure looks unrelated to this patch.

gstcefdemux.cc
```diff
@@ -83,13 +90,76 @@ typedef struct
...
+  return clock_time - base_time;
+}
+#endif
+
+
 static gboolean
 gst_cef_demux_push_audio_buffer (GstBuffer **buffer, guint idx, AudioPushData *push_data)
 {
-  GST_BUFFER_PTS (*buffer) += push_data->demux->ts_offset;
+  push_data->demux->last_audio_time = gst_element_get_current_running_time (GST_ELEMENT_CAST (push_data->demux));
```
reinismu, line 148: So what if GST version is 1.18? It seems like that method won't be defined then

philn: It should, see https://github.com/centricular/gstcefsrc/pull/60/commits/079ea274ddf97d0165e72665f90b114088bd7092

hevc:
I checked the latest patch with and without audio. Everything is working.Thanks.

MathieuDuponchelle:
lgtm, let's go :)

PR #61, opened 2022-12-13
CI: Install libunwind

philn:
Otherwise the Ubuntu 22.04 workflow fails with this error:

The following packages have unmet dependencies:
 libunwind-14-dev : Breaks: libunwind-dev but 1.3.2-2build2 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -17,6 +17,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


philn:
@MathieuDuponchelle when you have a minute to review this, please 🙏🏻 :)

MathieuDuponchelle:
Thank you very much :)

PR #63, opened 2022-12-16
Allow commas in chrome flags

frostbyte73:
Before: 
`gstcefsrc.SetProperty("chrome-extra-flags", "disable-features=site-per-process,TranslateUI,BlinkGenPropertyTrees")`
```
Adding switch with value disable-features=site-per-process
Adding flag TranslateUI
Adding flag BlinkGenPropertyTrees
```

After:
`gstcefsrc.SetProperty("chrome-extra-flags", "disable-features=site-per-process|TranslateUI|BlinkGenPropertyTrees")`
```
Adding switch with value disable-features=site-per-process,TranslateUI,BlinkGenPropertyTrees
```

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -410,8 +410,14 @@ class App : public CefApp
         gchar **switch_value = g_strsplit ((const gchar *) flags_list[i], "=", -1);
 
         if (g_strv_length (switch_value) > 1) {
-          GST_INFO_OBJECT (src, "Adding switch with value %s=%s", switch_value[0], switch_value[1]);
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
Ah, I see, that's a bit of a bummer :( Are there any guarantees that `|` will indeed not occur in values ?

frostbyte73:
> Ah, I see, that's a bit of a bummer :( Are there any guarantees that `|` will indeed not occur in values ?

No guarantees, but generally the list flags are comma separated. I considered trying to escape the commas instead but my c++ is pretty rusty

MathieuDuponchelle:
I see. Would turning the property into an array of strings instead solve the issue?

PR #64, opened 2022-12-16
Post console messages to bus

frostbyte73:
Allows the pipeline to react to your web app through console messages

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -259,27 +259,44 @@ public:
 
   virtual bool OnConsoleMessage(CefRefPtr<CefBrowser>, cef_log_severity_t level, const CefString &message, const CefString &source, int line) override {
     GstDebugLevel gst_level = GST_LEVEL_NONE;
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
oof, that is extremely spammy, you're not really supposed to post logs on the bus. I suppose applications could instead install a custom log handler to achieve the same results, but in reality logs are not supposed to be API, surely there are better ways to achieve your particular use case :)

frostbyte73:
> oof, that is extremely spammy, you're not really supposed to post logs on the bus. I suppose applications could instead install a custom log handler to achieve the same results, but in reality logs are not supposed to be API, surely there are better ways to achieve your particular use case :)

Ah, makes sense. What we really want is to start and stop the pipeline according to certain log messages, and also a way to pass logs to our application logger instead of printing to stdout. I'll look into using a custom log handler

MathieuDuponchelle:
@frostbyte73 OK, I'll close this then but don't hesitate to ask if you have issues, and thanks for taking the time to propose a patch :)

Issue #66, opened 2023-01-20
Not compiling on macOS Ventura on MacBook M1


pldin601:
Can't compile the plugin on MacBook Pro M1 (macOS Ventura):

Here's full build cast:
[![asciicast](https://asciinema.org/a/tqhIPm4XpeiG5ljTF3imhH0Ab.svg)](https://asciinema.org/a/tqhIPm4XpeiG5ljTF3imhH0Ab)

Have all gstreamer-related dependencies installed. During cmake run, I saw this warning several times:
```
IMPORTED_LOCATION not set for imported target "libcef_lib" configuration
  "Release".
```

UPD: Found possible reason. This build https://cef-builds.spotifycdn.com/index.html#macosarm64 has only sandboxed version of module `cef_sandbox.a` and `CMakeLists.txt` has no support if CEF sandboxed. Because in that case, env with paths to that module have a bit different names.

PR #67, opened 2023-01-20
fix builds on macos arm64 and pay attention to USE_SANDBOX flag

pldin601:
Fixes https://github.com/centricular/gstcefsrc/issues/66


reinismu:
And it actually works? From what I recall it couldn't  launch cef properly. Started to look into it a while ago https://github.com/centricular/gstcefsrc/compare/master...reinismu:gstcefsrc:macos-support Tho didn't get far

pldin601:
@reinismu At least, it started compiling, but `gstcefsubprocess` runs into segmentation fault...

MathieuDuponchelle:
Hah, thanks for the patch I suppose, I will merge it because you say it makes things compile when they didn't before, but I don't really have a clue what we're doing here :) Good luck to you on making this actually work beyond compiling though, I hope you figure it out :)

Issue #68, opened 2023-01-24
Broken Linux x64 build


SteveMcFarlin:
The following PR broke the x64 Linux build: https://github.com/centricular/gstcefsrc/pull/67

```
[ 96%] Building CXX object CMakeFiles/gstcefsubprocess.dir/gstcefsubprocess.cc.o
make[2]: *** No rule to make target 'libcef_lib-NOTFOUND', needed by 'Release/gstcefsubprocess'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:102: CMakeFiles/gstcefsubprocess.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

Reverting to the prior commit fixes the build.

philn:
cc @pldin601

MathieuDuponchelle:
Sorry about that, closing, thanks for reporting!

PR #70, opened 2023-01-30
Revert "fix builds on macos arm64 and pay attention to USE_SANDBOX flag"

pldin601:
Reverts centricular/gstcefsrc#67 due to the issue #68


MathieuDuponchelle:
Sorry, I was away on business trip, thanks, I guess there's not much harm done in merging this as the patch didn't make macOS actually  work?

Issue #73, opened 2023-06-03
Bad audio meta


heftig:
Current master (79fbffc) makes our app explode:

```
GStreamer-Audio: gst_audio_buffer_map: assertion 'meta->samples <=
  gst_buffer_get_size (gstbuffer) / GST_AUDIO_INFO_BPF (&meta->info)' failed
```

I think I can spot some issues:

- `audio_info` isn't updated when the caps change.
- The buffer size is passed to `gst_buffer_add_audio_meta` instead of the sample count.

MathieuDuponchelle:
While 1. is possible in theory, I don't think it can happen with `cef` (though it can't hurt to check that). 2. is indeed just wrong afaict, @philn any reason for just doing:

```
gst_buffer_add_audio_meta (*buffer, &push_data->demux->audio_info, gst_buffer_get_size (*buffer), NULL);
```
I reckon here we want `gst_buffer_get_size (*buffer) / push_data->demux->audio_info.bpf * push_data->demux->audio_info.channels` , can you try that @heftig ?

MathieuDuponchelle:
or `gst_buffer_get_size (*buffer) / GST_AUDIO_INFO_BPS (&push_data->demux->audio_info)`, should be equivalent

philn:
omg what was I thinking... yeah this code i wrote is clearly wrong.

philn:
I guess this can be closed then! Next time a "Fixes #issuenumber" in commit message should do it automatically. Thanks @SteveMcFarlin !

heftig:
This is still broken as far as I can tell. It should be using `BPF` and not `BPS`.

`gst_buffer_add_audio_meta` expects the number of samples in the buffer. The wording is a bit ambiguous, but the example given for non-interleaved layout makes it clear that the buffer size should be equal to `channels * samples * sample_stride`.

`BPS` produces the bytes per sample (the `sample_stride`), while `BPF` produces the bytes per frame (`channels * sample_stride`).

So in the case of a non-mono stream I'd expect the meta to still be broken.

MathieuDuponchelle:
(@heftig ^)

heftig:
In the context of audio-info, a frame is defined as one sample for each channel. Now look at the way the way `gst_ffmpegauddec_audio_frame` adds the meta:

```c
nsamples = ffmpegdec->frame->nb_samples;
```

```c
[...]

PR #75, opened 2023-08-02
Fix bad audio metadata (issues #73)

SteveMcFarlin:
It looks like issue #73 was never addressed. This PR has the suggested fix.

### DIFF
--- a/gstcefdemux.cc
+++ b/gstcefdemux.cc
@@ -149,7 +149,9 @@ gst_cef_demux_push_audio_buffer (GstBuffer **buffer, guint idx, AudioPushData *p
   GST_BUFFER_DTS (*buffer) = push_data->demux->last_audio_time;
   GST_BUFFER_PTS (*buffer) = push_data->demux->last_audio_time;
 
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
lgtm thanks :)

heftig:
Shouldn't this be using `BPF` instead of `BPS`?

SteveMcFarlin:
@heftig I am not sure of the answer here. You might want to open an issue if you think BPF is the right choice.

MathieuDuponchelle:
Why would `BPF` be correct here? The third parameter to `gst_buffer_add_audio_meta` is a number of samples no?

heftig:
`gst_buffer_add_audio_meta` expects the number of samples in the buffer. The wording is a bit ambiguous, but the example given for non-interleaved layout makes it clear that the buffer size should be equal to `channels * samples * sample_stride`.

`BPS` produces the bytes per sample (the `sample_stride`), while `BPF` produces the bytes per frame (`channels * sample_stride`).

So in the case of a non-mono stream I'd expect the meta to still be broken.

MathieuDuponchelle:
Please see my answer [here](https://github.com/centricular/gstcefsrc/issues/73#issuecomment-1701008590)

Issue #29, opened 2021-06-11
[question] h264 CEF?


ldenoue:
How could we use a build of CEF that has all codecs, like H264?

I noticed this commit changin remote port numbers, but is there one to tell CEF to use a different Chromium build? Or has CEF to be built completely from scratch?

MathieuDuponchelle:
you need to build CEF from scratch I'm afraid:

https://stackoverflow.com/questions/64293697/enable-aac-support-in-chromium-embedded-framework-c

Good luck because it's an absolute PITA :P

reinismu:
It's not so bad, I managed to do it in a day. I could link to binaries, but it is in my S3 and I don't want to pay for aws too much :D 

@ldenoue Here is the docker and scripts how to do it. Should be pretty simple to do

MathieuDuponchelle:
@reinismu did you forget to link your scripts?

reinismu:
> @reinismu did you forget to link your scripts?

Yup :D Added link

ldenoue:
@reinismu thanks a lot! I'll try your scripts.
Laurent

Issue #25, opened 2021-05-20
Gst Demux Audio Caps


Townsheriff:
Hey,

I wanted to say big thank you for this project.

On the side note. I'was building pipeline with decklink video and audio sinks which are taken from the cefsrc plugin, but I found an issue with audio caps, in the plugin inspect and in the dot fiels I see format to be F32BE, but when I actually add it in the caps it shows me that it failed to negotiate.

Some code:
```
    cef_audio_filter.set_property("caps", &gst::Caps::builder("audio/x-raw")
        // here is a bug with invalid format, shows that it is F32BE, but when added in the caps fails to negotiate
        // .field("format", &"F32BE")
        .field("rate", &44100)
        .field("channels", &2)
        .field("layout", &"interleaved")
        .build()
    ).unwrap();
```

Some logs:
```
0:00:05.547767871 903835 0x561f142ca180 WARN                 basesrc gstbasesrc.c:3072:gst_base_src_loop:<cefsrc0> error: Internal data stream error.
0:00:05.547796675 903835 0x561f142ca180 WARN                 basesrc gstbasesrc.c:3072:gst_base_src_loop:<cefsrc0> error: streaming stopped, reason not-negotiated (-4)
Error received Some("/GstPipeline:pipeline0/GstCefSrc:cefsrc0"): Internal data stream error.
```

If I don't add the field, it works fine.

MathieuDuponchelle:
Indeed, the advertised F32BE format in the template caps is wrong, and the final caps that got pushed were F32LE, leading to the negotiation problem. I just merged https://github.com/centricular/gstcefsrc/pull/18 , with that you can specify the correct format (F32LE) in the filter. Thanks for the report :)

Issue #24, opened 2021-05-11
always in "Setting pipeline to PAUSED ..."


ldenoue:
Running on Ubuntu any sample command gets me "Setting pipeline to PAUSED ...".
Any idea why?
Thank you!

MathieuDuponchelle:
Afraid not, what sample command were you using?

ldenoue:
I'm using the sample command from the README.md:

```
GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 \
    cefsrc url="https://soundcloud.com/platform/sama" ! \
    video/x-raw, width=1920, height=1080, framerate=60/1 ! \
    cefdemux name=demux ! videoconvert ! \
    queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! x264enc ! \
    mp4mux name=muxer fragment-duration=1000 ! filesink location='test.mp4' \
    audiotestsrc do-timestamp=true is-live=true  volume=0.00 ! audiomixer name=mix ! \
    queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! audioconvert ! \
    audiorate ! audioresample ! voaacenc bitrate=128000 ! muxer. \
    demux. ! mix.
```

I also tried this one:
```
GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 cefsrc url="https://google.com" !     video/x-raw, width=1920, height=1080, framerate=60/1 !     cefdemux name=demux ! videoconvert ! queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! x264enc !     mp4mux name=muxer fragment-duration=1000 ! filesink location='test.mp4'     demux. ! queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! audioconvert ! audiorate ! audioresample ! voaacenc bitrate=128000 ! muxer.
```

but both simply display:
"Setting pipeline to PAUSED ..."

Is it maybe the CEF version that somehow doesn't fire "page loaded" or something?

Notice I had to use vocaacenc instead of faac because I couldn't find a way to install faac plugin on my Ubuntu: if it could be the reason why the pipeline is stuck to "PAUSED", would you have a simpler pipeline sample that saves to .webm maybe with an opusenc?

ldenoue:
[...]

PR #17, opened 2021-03-02
Add website record example

reinismu:
This is how I wrote it. It might not be the best way. Let me know if there is anything to improve in my pipeline :)

PR #83, opened 2024-03-27
Reland "Fix build on macosarm64 and pay attention to USE_SANDBOX flag"

SteveMcFarlin:
The issue with the prior PR was USE_SANBOX was ON by default. I clearly did not pay attention enough to that. IMO the default should be OFF, and only be enabled if the arch is arm and the target is macos. 

- This updates the CEF version to 122.1.13. The reason for this is on the M1 you will get a flood of linker errors due to incompatibilities with the older 103.0.9.
- I had to rename USE_SANBOX to SANDBOX as find_packages was resetting it to ON if set to OFF prior to the call
- Tested build/inspect on Mac M1 in Release and Linux 64 in Release

Additionally I had to add a few additional cmake options for the MacARM build otherwise I would get errors, and in one case lib_cef_dllwrapper was being compiled for i386. Let me know if there is something wrong there.

Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin23.2.0
cmake version 3.24.1


reinismu:
It worked on MacOS as well or just compiled?

SteveMcFarlin:
I compiled and did a gst-inspect-1.0 on it. I did not test but I can run one now... BUMMER. SIGSEGV. Changing to draft. I will work on it tomorrow. It's late for me.

SteveMcFarlin:
`0:00:00.789074000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:2194:gst_aggregator_query_latency_unlocked:<mix> Signaling src from thread 0x60000396f2a0
0:00:00.789077000 57748 0x60000396f2a0 DEBUG             aggregator gstaggregator.c:2197:gst_aggregator_query_latency_unlocked:<mix> configured latency live:true min:33219954 max:-1
0:00:00.789096000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:488:gst_aggregator_check_pads_ready:<mix> checking pads
0:00:00.789099000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:532:gst_aggregator_check_pads_ready:<mix:sink_0> Have no buffer and not EOS yet
0:00:00.789101000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:532:gst_aggregator_check_pads_ready:<mix:sink_1> Have no buffer and not EOS yet
0:00:00.789103000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:587:gst_aggregator_check_pads_ready:<mix> pad not ready to be aggregated yet
0:00:00.789106000 57748 0x60000396f2a0 LOG               aggregator gstaggregator.c:880:gst_aggregator_wait_and_check:<mix> Waiting for src on thread 0x60000396f2a0
* thread #1, stop reason = signal SIGSTOP`

reinismu:
You can see here https://github.com/centricular/gstcefsrc/pull/67 Plus I started on my branch to make it work. https://github.com/centricular/gstcefsrc/compare/master...reinismu:gstcefsrc:macos-support but I'm not familiar with cmake and so so with MacOS. 

We might put a bounty on this thing. Running this plugin on MacOS would benefit us.

reinismu:
We had a discussion and we decided to reward **1000$** to someone who can make this plugin work on Apple Silicon MacOS :)

SteveMcFarlin:
You are headed on the right path with what you have done in gstcefsubprocess. If If I were to do more work on this I would look to the CEF examples. Specifically the mac [process](https://github.com/chromiumembedded/cef-project/blob/master/examples/shared/process_helper_mac.cc) helper. If I have time this evening I will see what I can come up with.

SteveMcFarlin:
So over lunch I took another look at this. The segfault is coming from CefString. In `static gpointer run_cef(GstCefSrc *src) {` There is the line: 

```
CefString(&settings.browser_subprocess_path)
      .FromASCII(browser_subprocess_path);
```

That is where the segfault happens. I will look into it more later today if I have time.

SteveMcFarlin:
At this point not sure what is going on here.  From the code `CefString str("LETS CRASH");`.

```
(lldb) bt
* thread #17, name = 'cef-ui-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x0000000104291d84 libgstcef.dylib`cef_string_utf8_to_utf16(src="LETS CRASH", src_len=10, output=0x0000600003fc6500) at libcef_dll_dylib.cc:1583:10
    frame #2: 0x00000001041a6f60 libgstcef.dylib`CefStringTraitsUTF16::from_string(data="LETS CRASH", length=10, s=0x0000600003fc6500) at cef_string_wrappers.h:269:12
    frame #3: 0x00000001041a6e38 libgstcef.dylib`CefStringBase<CefStringTraitsUTF16>::FromString(this=0x00000001706babe0, data="LETS CRASH", length=10) at cef_string_wrappers.h:684:12
    frame #4: 0x00000001041a6d6c libgstcef.dylib`CefStringBase<CefStringTraitsUTF16>::CefStringBase(this=0x00000001706babe0, src="LETS CRASH", length=0) at cef_string_wrappers.h:385:7
    frame #5: 0x00000001041a61f8 libgstcef.dylib`CefStringBase<CefStringTraitsUTF16>::CefStringBase(this=0x00000001706babe0, src="LETS CRASH", length=0) at cef_string_wrappers.h:383:38
    frame #6: 0x00000001041ab128 libgstcef.dylib`run_cef(src=0x000000013c0682e0) at gstcefsrc.cc:474:13
    frame #7: 0x00000001007101fc libglib-2.0.0.dylib`g_thread_proxy + 68
    frame #8: 0x0000000188bb6034 libsystem_pthread.dylib`_pthread_start + 136
```

SteveMcFarlin:
I thought maybe trying to load the library in gstcefsrc.c might be the issue, but as soon as I touch 
```
// Initialize the macOS sandbox for this helper process.
  CefScopedSandboxContext sandbox_context;
  ```
I will get a crash. I think this might be a mem allocator issue, or __possibly_ a mismatch with libstdc++? CefString does an allocation, and so does CefScopedSandboxContext. I can do this 
```
CefBrowserSettings browserSettings;

  settings.no_sandbox = !src->sandbox;
  settings.windowless_rendering_enabled = true;
  settings.log_severity = src->log_severity;
 ```
which is fine given this is a simple structure on the stack. It has been a while since I have dealt with these types of issues. I guess if I were going to debug this further I would build CEF against my local tool chain. At the very least I could generate a debug build.

SteveMcFarlin:
Closing this as I will not have any more time to tinker with this.

SteveMcFarlin:
@reinismu I would like to get this working at some point, but I simply do not have the time to dig into it. As I stated above the issue revolves around memory allocation AFAICT. There is a bad access at as you can see. I ran the CEF sample minimal without any issue, so I know CEF works on the OS-X and the M series chips. At this point I don't know. I have a habit of closing PR's that I will not work on. However, given minimal amount of traffic here I will reopen in hopes someone will have an aha moment.

reinismu:
Maybe someone from Centricular will get interested in this as well.

I wonder if bad access is not caused by MacOS security stuff. The last I recall from my try was setting up process metadata

SteveMcFarlin:
I am compiling CEF locally, so I can at least debug a bit better. I will continue to look into this when I have some free time.

MathieuDuponchelle:
> We had a discussion and we decided to reward **1000$** to someone who can make this plugin work on Apple Silicon MacOS :)

Noted, I will check with our MacOS people :)

reinismu:
> > We had a discussion and we decided to reward **1000$** to someone who can make this plugin work on Apple Silicon MacOS :)
> 
> Noted, I will check with our MacOS people :)

Did any MacOS people seem interested? :) If price is too low, we can discuss it

amyspark:
Hi! @nirbheek told me I could have a look at this. The issue you were facing is that, on macOS, what you're calling is in fact an interposer, lazy-loader dylib: https://github.com/chromiumembedded/cef/blob/00118ddcdbdd6f7a5ed6e131417e6fb222bb33b0/libcef_dll/wrapper/cef_library_loader_mac.mm#L63-L65

Since that library was never initialized in the gstcef side, all CEF calls were rendered null pointer dereferences. I believe these changes should be sufficient to get you going, feel free to include and rebase them : https://github.com/SteveMcFarlin/gstcefsrc/compare/cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64

Bear in mind, however, that it seems CEF intends you to consume the library as an Apple framework -- the loader attempts to find the Framework relative to the executable's path, as when deployed into an app bundle.

reinismu:
> Hi! @nirbheek told me I could have a look at this. The issue you were facing is that, on macOS, what you're calling is in fact an interposer, lazy-loader dylib: https://github.com/chromiumembedded/cef/blob/00118ddcdbdd6f7a5ed6e131417e6fb222bb33b0/libcef_dll/wrapper/cef_library_loader_mac.mm#L63-L65
> 
> Since that library was never initialized in the gstcef side, all CEF calls were rendered null pointer dereferences. I believe these changes should be sufficient to get you going, feel free to include and rebase them : [SteveMcFarlin/gstcefsrc@cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64](https://github.com/SteveMcFarlin/gstcefsrc/compare/cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64)
> 
> Bear in mind, however, that it seems CEF intends you to consume the library as an Apple framework -- the loader attempts to find the Framework relative to the executable's path, as when deployed into an app bundle.

Thanks for looking into this :)

Currently trying it out on my M2 Mac Mini. Compiled and set up everything on it.

Sadly for me it segfaults :(

Command
```
GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 -e \
    cefsrc url="https://soundcloud.com/platform/sama" ! \
    video/x-raw, width=1920, height=1080, framerate=60/1 ! \
    cefdemux name=demux ! queue ! videoconvert ! \
    queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! x264enc ! queue ! \
    mp4mux name=muxer ! filesink location='test.mp4' \
    audiotestsrc do-timestamp=true is-live=true  volume=0.0 ! audiomixer name=mix ! \
    queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! audioconvert ! \
    audiorate ! audioresample ! faac bitrate=128000 ! queue ! muxer. \
    demux. ! queue ! mix.
```

Logs:
```
.... setting up pipeline

0:00:00.026971459 30738 0x600000391770 INFO               structure gststructure.c:2957:gst_structure_get_valist: Expected field 'channel-mask' in structure: audio/x-raw, rate=(int)[ 1, 2147483647 ], channels=(int)[ 1, 2147483647 ];
0:00:00.026979917 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2825:gst_element_continue_state:<queue5> completed state change to PAUSED
0:00:00.026985000 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2728:_priv_gst_element_state_changed:<queue5> notifying about state-changed READY to PAUSED (VOID_PENDING pending)
0:00:00.026989375 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2942:gst_bin_change_state_func:<pipeline0> child 'queue5' changed state to 3(PAUSED) successfully
0:00:00.026992834 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2484:gst_bin_element_set_state:<demux> current READY pending VOID_PENDING, desired next PAUSED
0:00:00.026996125 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2825:gst_element_continue_state:<demux> completed state change to PAUSED
0:00:00.026998625 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2728:_priv_gst_element_state_changed:<demux> notifying about state-changed READY to PAUSED (VOID_PENDING pending)
0:00:00.027001792 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2942:gst_bin_change_state_func:<pipeline0> child 'demux' changed state to 3(PAUSED) successfully
0:00:00.027004542 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2484:gst_bin_element_set_state:<capsfilter0> current READY pending VOID_PENDING, desired next PAUSED
0:00:00.027007459 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2825:gst_element_continue_state:<capsfilter0> completed state change to PAUSED
0:00:00.027009834 30738 0x6000003ac370 INFO              GST_STATES gstelement.c:2728:_priv_gst_element_state_changed:<capsfilter0> notifying about state-changed READY to PAUSED (VOID_PENDING pending)
0:00:00.027012709 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2942:gst_bin_change_state_func:<pipeline0> child 'capsfilter0' changed state to 3(PAUSED) successfully
0:00:00.027015250 30738 0x6000003ac370 INFO              GST_STATES gstbin.c:2484:gst_bin_element_set_state:<cefsrc0> current READY pending VOID_PENDING, desired next PAUSED
0:00:00.027054750 30738 0x6000003a00a0 WARN              aggregator gstaggregator.c:2283:gst_aggregator_query_latency_unlocked:<mix> Latency query failed
0:00:00.027065417 30738 0x6000003ae620 INFO                  cefsrc gstcefsrc.cc:481:run_cef: Initializing CEF
zsh: segmentation fault  GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 -e cefsrc  !     !
```

I tried running `gstcefsubprocess` manually and it segfaults as well.

```
sharehand@SH-MacMini-M2 Release % sudo lldb ./gstcefsubprocess
(lldb) target create "./gstcefsubprocess"
Current executable set to '/Users/sharehand/projects/gstcefsrc/build/Release/gstcefsubprocess' (arm64).
(lldb) run
Process 30799 launched: '/Users/sharehand/projects/gstcefsrc/build/Release/gstcefsubprocess' (arm64)
Process 30799 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000000000000
error: memory read failed for 0x0
Target 0: (gstcefsubprocess) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x0000000100036f50 gstcefsubprocess`CefExecuteProcess(CefMainArgs const&, scoped_refptr<CefApp>, void*) + 36
    frame #2: 0x000000010000278c gstcefsubprocess`main + 120
    frame #3: 0x000000019547bf28 dyld`start + 2236
(lldb)
```

Maybe I missed something?

amyspark:
Hi! That's correct, CEF libraries are missing because on macOS, *there are none*. What the Spotify tarball ships is instead a framework.

I assume the intention of the developers was to consume it from an app, and so they architected the lazy load to resolve from said path (`path_to_executable/../Frameworks/Chromium Embedded Framework.framework/...`). It'll fail completely if you try to use it from a standalone executable, are you building such an example?

reinismu:
> Hi! That's correct, CEF libraries are missing because on macOS, _there are none_. What the Spotify tarball ships is instead a framework.
> 
> I assume the intention of the developers was to consume it from an app, and so they architected the lazy load to resolve from said path (`path_to_executable/../Frameworks/Chromium Embedded Framework.framework/...`). It'll fail completely if you try to use it from a standalone executable, are you building such an example?

Did you manage to run gstcefsrc on your end? If so could you provide example?

SteveMcFarlin:
> Hi! @nirbheek told me I could have a look at this. The issue you were facing is that, on macOS, what you're calling is in fact an interposer, lazy-loader dylib: https://github.com/chromiumembedded/cef/blob/00118ddcdbdd6f7a5ed6e131417e6fb222bb33b0/libcef_dll/wrapper/cef_library_loader_mac.mm#L63-L65
> 
> Since that library was never initialized in the gstcef side, all CEF calls were rendered null pointer dereferences. I believe these changes should be sufficient to get you going, feel free to include and rebase them : [SteveMcFarlin/gstcefsrc@cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64](https://github.com/SteveMcFarlin/gstcefsrc/compare/cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64)
> 
> Bear in mind, however, that it seems CEF intends you to consume the library as an Apple framework -- the loader attempts to find the Framework relative to the executable's path, as when deployed into an app bundle.

I will rebase this later this week and mark the PR as open. Unfortunately I am unable to test this as my M1 was wiped during a layoff, and I have not had a chance to get my CEF/GStreamer M1 dev environment back up. Thanks for looking into this.

SteveMcFarlin:
@reinismu I have not yet. I am getting setup to test it either today or tomorrow.

SteveMcFarlin:
@reinismu I have finally got a system up and running. I installed GStreamer from the downloads. I can not even load the plugin currently. 

```
dyld[53231]: Library not loaded: @rpath/libgstreamer-1.0.0.dylib
  Referenced from: <07B6FD71-26C5-37F7-9B7A-DC944038F101> /Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/gst-plugin-scanner
  Reason: tried: '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/lib/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../lib/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../../libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/lib/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../lib/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libgstreamer-1.0.0.dylib' (no such file), '/Library/Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/../../libgstreamer-1.0.0.dylib' (no such file), '/usr/local/lib/libgstreamer-1.0.0.dylib' (no such file), '/usr/lib/libgstreamer-1.0.0.dylib' (no such file, not in dyld cache)
  ```

Additionally I get errors related to `../Frameworks/Chromium Embedded Framework.framework/Chromium Embedded Framework`. Prior to trying the installers I built from source, and received the same errors. I may try a home-brew installation, but for now I am stuck.

**Comment by SteveMcFarlin** [comment]
I am now only having issues loading the CEF Framework with the distro from Spotify. I am not used to using those packages, so I am build CEF locally in order to see if I can get something working. I currently have no intention of building an App via the usual methods on MacOS, so I am going to work on that.

**Comment by SteveMcFarlin** [comment]
@reinismu I had to copy the framework into my gst installation. Now I get a crash. I have a feeling this is due to issues revolving around CEF not being bundled in an app as Amy pointed out. 

```
Setting pipeline to PAUSED ...
Process 54332 stopped
* thread #18, name = 'cef-ui-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x147ef4020)
    frame #0: 0x0000000147ef4020 Chromium Embedded Framework`___lldb_unnamed_symbol7480 + 600424
Chromium Embedded Framework`___lldb_unnamed_symbol7480:
->  0x147ef4020 <+600424>: brk    #0
    0x147ef4024 <+600428>: hlt    #0
    0x147ef4028 <+600432>: brk    #0x1
    0x147ef402c <+600436>: mov    w0, #0x2
Target 0: (gst-launch-1.0) stopped.
(lldb) bt
* thread #18, name = 'cef-ui-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x147ef4020)
  * frame #0: 0x0000000147ef4020 Chromium Embedded Framework`___lldb_unnamed_symbol7480 + 600424
    frame #1: 0x000000014769171c Chromium Embedded Framework`___lldb_unnamed_symbol7466 + 2518368
    frame #2: 0x0000000147690a68 Chromium Embedded Framework`___lldb_unnamed_symbol7466 + 2515116
    frame #3: 0x000000014459f39c Chromium Embedded Framework`cxxbridge1$rust_vec$u8$set_len + 216008
    frame #4: 0x000000014459f114 Chromium Embedded Framework`cxxbridge1$rust_vec$u8$set_len + 215360
    frame #5: 0x000000014457cd50 Chromium Embedded Framework`cxxbridge1$rust_vec$u8$set_len + 75132
    frame #6: 0x000000014457ca9c Chromium Embedded Framework`cxxbridge1$rust_vec$u8$set_len + 74440
    frame #7: 0x00000001444df0e8 Chromium Embedded Framework`cef_initialize + 312
    frame #8: 0x0000000104025230 libgstcef.dylib`CefInitialize(CefMainArgs const&, CefStructBase<CefSettingsTraits> const&, scoped_refptr<CefApp>, void*) + 184
    frame #9: 0x0000000103fedb60 libgstcef.dylib`run_cef(_GstCefSrc*) + 792
    frame #10: 0x0000000100957c7c libglib-2.0.0.dylib`g_thread_proxy + 68
    frame #11: 0x000000019d73af94 libsystem_pthread.dylib`_pthread_start + 136
```

**Comment by SteveMcFarlin** [comment]
Well I give up for now. I spent the day on it. I went so far as to create a console app and link everything properly. Still crashes when cefsrc initializes Cef.

**Comment by reinismu** [comment]
> Well I give up for now. I spent the day on it. I went so far as to create a console app and link everything properly. Still crashes when cefsrc initializes Cef.

Thanks! Currently contracting Centricular (Amy) to try and solve it :) Now I just hope that issue is not too complicated

**Comment by SteveMcFarlin** [comment]
I am still gong to do a local CEF build the way I do it on linux to see if it is any different. Hopefully this does get fixed. Always wanted it working on OS-X arm64 just for convenience.

**Comment by amyspark** [comment]
Hi again @SteveMcFarlin @reinismu !

The reason why the update has been so troublesome on macOS is because Chromium's Sandbox V2 goes out of its way to ensure that the framework is loaded from within an app bundle. So far I have discovered the following:

- There is no workaround for initializing the plugin from inside an app bundle, Chromium itself enforces that at initialization
- There are overrides needed to allow loading the CEF framework itself from outside that app bundle, for ad-hoc/development use
- When the Cocoa event loop is not managed by CEF itself (see the call to `CefRunMessageLoop()` in the existing gstcefsrc.cc code) its callee *must* be aware of CEF and implement a protocol in the `NSApplication` rendering the UI

For the first item, I put together an app bundle for `gst-launch-1.0` by moving the executable to `bin/gst-launch-1.0.app/Contents/MacOS/gst-launch-1.0`, symlinking it back to the `bin` folder, and adding an Info.plist. The complete bundle should look like this: [gst-launch-1.0.zip](https://github.com/user-attachments/files/15820616/gst-launch-1.0.zip)

For the second item, the overrides are in detail here: https://github.com/amyspark/gstcefsrc/commit/9930d07044c56e895e9fa91b154849444787ddf9

For the third item, I started working on that ([commit here](https://github.com/amyspark/gstcefsrc/commit/0df4600751eef311baff09dccb70f2afd9b639c1), but I still need to port the following in the Java bindings: https://github.com/chromiumembedded/java-cef/blob/master/native/util_mac.mm#L70-L73).

I'll keep you posted, you can find what progress I make in [this branch](https://github.com/SteveMcFarlin/gstcefsrc/compare/cmake-macosarm64...amyspark:gstcefsrc:cmake-macosarm64).

**Comment by reinismu** [comment]
Hi @amyspark !

Thanks for your work :)

Tried out your latest branch changes and noticed few issues

1) Couldn't compile it at start

removed as it seemed to not do anything anyway and it worked
```
-      if (!init_result) {
-        // BAIL OUT, CEF is not loaded.
-        result = GST_STATE_CHANGE_FAILURE;
-        g_thread_join(thread);
-        thread = nullptr;
-      }
```

2) I'm not sure what you mean by 

>For the first item, I put together an app bundle for gst-launch-1.0 by moving the executable to bin/gst-launch-1.0.app/Contents/MacOS/gst-launch-1.0, symlinking it back to the bin folder, and adding an Info.plist. The complete bundle should look like this: [gst-launch-1.0.zip](https://github.com/user-attachments/files/15820616/gst-launch-1.0.zip)

I assume that is the reason why it still doesnt work for me

```
0:00:14.343820209 96241 0x6000003caa30 INFO                  cefsrc gstcefsrc.cc:640:run_cef:<cefsrc0> Initialize2 CEF
warning: Chromium Embedded Framework was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 96241 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x111e57fd8)
    frame #0: 0x0000000111e57fd8 Chromium Embedded Framework`base::apple::(anonymous namespace)::BundleFromPath(base::FilePath const&) [inlined] base::ImmediateCrash() at immediate_crash.h:176:3 [opt]
Target 0: (gst-launch-1.0) stopped.
Process 96241 launched: '/opt/homebrew/bin/gst-launch-1.0' (arm64)
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x111e57fd8)
  * frame #0: 0x0000000111e57fd8 Chromium Embedded Framework`base::apple::(anonymous namespace)::BundleFromPath(base::FilePath const&) [inlined] base::ImmediateCrash() at immediate_crash.h:176:3 [opt]
    frame #1: 0x0000000111e57fd8 Chromium Embedded Framework`base::apple::(anonymous namespace)::BundleFromPath(base::FilePath const&) [inlined] logging::CheckFailure() at check.h:211:3 [opt]
    frame #2: 0x0000000111e57fd8 Chromium Embedded Framework`base::apple::(anonymous namespace)::BundleFromPath(file_path=<unavailable>) at bundle_locations.mm:66:3 [opt]
    frame #3: 0x0000000111e57f38 Chromium Embedded Framework`base::apple::SetOverrideOuterBundlePath(file_path=<unavailable>) at bundle_locations.mm:82:22 [opt]
    frame #4: 0x000000010e69f808 Chromium Embedded Framework`util_mac::BasicStartupComplete() [inlined] util_mac::(anonymous namespace)::OverrideOuterBundlePath() at util_mac.mm:45:3 [opt]
    frame #5: 0x000000010e69f7f0 Chromium Embedded Framework`util_mac::BasicStartupComplete() at util_mac.mm:159:3 [opt]
    frame #6: 0x000000010e6598e4 Chromium Embedded Framework`AlloyMainDelegate::BasicStartupComplete(this=<unavailable>) at alloy_main_delegate.cc:509:3 [opt]
    frame #7: 0x00000001116dd668 Chromium Embedded Framework`content::ContentMainRunnerImpl::Initialize(this=0x00006000023e4a80, params=<unavailable>) at content_main_runner_impl.cc:875:18 [opt]
    frame #8: 0x00000001116dc86c Chromium Embedded Framework`content::ContentMainInitialize(params=<unavailable>, content_main_runner=<unavailable>) at content_main.cc:291:38 [opt]
    frame #9: 0x000000010e5e0a2c Chromium Embedded Framework`CefMainRunner::ContentMainInitialize(this=0x00006000018f0840, args=0x000000016fdfd9b0, windows_sandbox_info=<unavailable>, no_sandbox=<unavailable>) at main_runner.cc:422:10 [opt]
    frame #10: 0x000000010e5e079c Chromium Embedded Framework`CefMainRunner::Initialize(this=0x00006000018f0840, settings=0x0000000132e1c438, application=(ptr_ = 0x0000600000dcefb0), args=0x000000016fdfd9b0, windows_sandbox_info=<unavailable>, initialized=0x0000000132e1c430, context_initialized=base::OnceClosure @ 0x000000016fdfd658) at main_runner.cc:249:7 [opt]
    frame #11: 0x000000010e5b7c08 Chromium Embedded Framework`CefContext::Initialize(this=0x0000000132e1c430, args=0x000000016fdfd9b0, settings=<unavailable>, application=(ptr_ = 0x0000600000dcefb0), windows_sandbox_info=0x0000000000000000) at context.cc:480:22 [opt]
    frame #12: 0x000000010e5b7964 Chromium Embedded Framework`CefInitialize(args=0x000000016fdfd9b0, settings=0x000000016fdfd7e0, application=(ptr_ = 0x0000600000dcefb0), windows_sandbox_info=0x0000000000000000) at context.cc:312:19 [opt]
    frame #13: 0x000000010e524700 Chromium Embedded Framework`cef_initialize(args=<unavailable>, settings=<unavailable>, application=<unavailable>, windows_sandbox_info=0x0000000000000000) at libcef_dll.cc:114:7 [opt]
    frame #14: 0x00000001037113f8 libgstcef.dylib`CefInitialize(CefMainArgs const&, CefStructBase<CefSettingsTraits> const&, scoped_refptr<CefApp>, void*) + 184
    frame #15: 0x00000001036da2c0 libgstcef.dylib`run_cef(_GstCefSrc*) + 1708
    frame #16: 0x000000018d0c4910 libdispatch.dylib`_dispatch_client_callout + 20
    frame #17: 0x000000018d0d2fa8 libdispatch.dylib`_dispatch_main_queue_drain + 984
    frame #18: 0x000000018d0d2bc0 libdispatch.dylib`_dispatch_main_queue_callback_4CF + 44
    frame #19: 0x000000018d394ecc CoreFoundation`__CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 16
    frame #20: 0x000000018d3527d0 CoreFoundation`__CFRunLoopRun + 1996
    frame #21: 0x000000018d3519ac CoreFoundation`CFRunLoopRunSpecific + 608
    frame #22: 0x0000000197900448 HIToolbox`RunCurrentEventLoopInMode + 292
    frame #23: 0x0000000197900284 HIToolbox`ReceiveNextEventCommon + 648
    frame #24: 0x00000001978fffdc HIToolbox`_BlockUntilNextEventMatchingListInModeWithFilter + 76
    frame #25: 0x0000000190b2e8a4 AppKit`_DPSNextEvent + 660
    frame #26: 0x0000000191308980 AppKit`-[NSApplication(NSEventRouting) _nextEventMatchingEventMask:untilDate:inMode:dequeue:] + 716
    frame #27: 0x0000000190b21d50 AppKit`-[NSApplication run] + 476
    frame #28: 0x000000010063b1a8 libgstreamer-1.0.0.dylib`run_main_with_nsapp + 180
    frame #29: 0x000000010063b200 libgstreamer-1.0.0.dylib`gst_macos_main + 36
    frame #30: 0x000000018cef50e0 dyld`start + 2360
```

**Comment by amyspark** [comment]
Hey! Yes, that's exactly the reason -- it won't run at all, as it turns out, because Chromium's runtime checks at multiple steps that it is being run from an app bundle. The easiest way to try it out is by converting `gst-launch-1.0` to such a bundle. Let me know if you need help with it and I'll attach the steps here.

I should be finishing the branch today, I'll file a PR against @SteveMcFarlin 's repo as soon as I have it ready.

**Comment by nirbheek** [comment]
> I should be finishing the branch today, I'll file a PR against @SteveMcFarlin 's repo as soon as I have it ready.

I've added you as a maintainer @amyspark so you can just push to this PR.

**Comment by reinismu** [comment]
Some guidance on how to run it as app would be appreciated :) 
I downloaded your zip and checked it, but wasn't sure what to do next. Not that familiar with MacOS

**Comment by reinismu** [comment]
Gave it another go and managed to make gst-launch-1.0.app run. It just spent 30 minutes poking every dylib it could find on disk and crashed :/
Could you update readme.md on how to use it in MacOS?

**Comment by amyspark** [comment]
@reinismu Sure, I'll try to take care of this today.

**Comment by amyspark** [comment]
@reinismu 

> It just spent 30 minutes poking every dylib it could find on disk and crashed :/

Was it a RPATH issue (`dlopen` followed by many `not found` errors on the console)? I've been working on those, would it be possible for you to check again with the installer package from [the latest CI build](https://gitlab.freedesktop.org/gstreamer/cerbero/-/jobs/60445336)?

**Comment by reinismu** [comment]
> @reinismu
> 
> > It just spent 30 minutes poking every dylib it could find on disk and crashed :/
> 
> Was it a RPATH issue (`dlopen` followed by many `not found` errors on the console)? I've been working on those, would it be possible for you to check again with the installer package from [the latest CI build](https://gitlab.freedesktop.org/gstreamer/cerbero/-/jobs/60445336)?

I made an MacOS app differently so that's why it didn't work as it should.

I managed to run CefSrc on Mac after following your instructions.

I had to fix `BOOL` import for compiler to work
```
gstcefsrc/gstcefnsapplication.h:22:1: error: unknown type name 'BOOL'
BOOL gst_is_main_thread();
```

improted
```
#include <objc/objc.h>
#include <stdbool.h>
```

**Comment by nirbheek** [comment]
@amyspark can we squash some of these commits before merging?

**Comment by MathieuDuponchelle** [comment]
Just checked, still builds and runs on Linux, annoyingly I can reproduce the choppy audio that Piotr was mentioning in private chats. That's both with this version and `main` so obviously not a blocker here, but that will require attention eventually :|

I will review further tomorrow but this looks quite ready from a quick look at it :)

File: CMakeLists.txt [hunk 1]
```diff
@@ -1,25 +1,50 @@
-cmake_minimum_required(VERSION 2.8.12.1)
+cmake_minimum_required(VERSION 3.10.2)
+
+# Crash and burn if IMPORTED_LOCATION is set to NOT-FOUND -- 3.1
+cmake_policy(SET CMP0111 NEW)
```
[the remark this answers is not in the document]
amyspark: Good catch on both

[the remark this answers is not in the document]
amyspark: Bumped from 2.8 to 3.10.2, and from that (included in the rebase) to 3.19.

amyspark:
> @amyspark can we squash some of these commits before merging?

Sure @nirbheek , I've pushed a cleaned up version. Let me know if it looks OK.

gstcefsrc.cc
```diff
@@ -21,14 +31,40 @@ GST_DEBUG_CATEGORY_STATIC (cef_console_debug);
...
-#define DEFAULT_LOG_SEVERITY LOGSEVERITY_DISABLE
+#define DEFAULT_LOG_SEVERITY LOGSEVERITY_DEFAULT
+#if !defined(__APPLE__) && defined(GST_CEF_USE_SANDBOX)
+#define DEFAULT_SANDBOX TRUE
+#else
 #define DEFAULT_SANDBOX FALSE
+#endif
 
+// Blocks other elements from initializing CEF is it's already in progress.
+static gboolean cef_initializing = FALSE;
```
MathieuDuponchelle, line 42: Turning these three variables into an enum could make this nicer:

```
typedef enum {
    CEF_STATUS_INITIAL,
    CEF_STATUS_INITIALIZING,
    CEF_STATUS_INITIALIZED,
    CEF_STATUS_SHUTTING_DOWN,
} CefStatus;
```

What do you think?

gstcefsrc.cc
```diff
@@ -177,7 +213,7 @@ class AudioHandler : public CefAudioHandler
     GstStructure *s = gst_structure_new ("cef-audio-stream-start",
         "channels", G_TYPE_INT, channels,
         "rate", G_TYPE_INT, params.sample_rate,
-        NULL);
+        nullptr);
```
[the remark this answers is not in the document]
MathieuDuponchelle: oh clang doesn't accept NULL as a sentinel ? that's dumb :facepalm:

gstcefsrc.cc
```diff
@@ -340,8 +375,17 @@ void BrowserClient::OnBeforeClose(CefRefPtr<CefBrowser> browser) {
   mElement->browser = nullptr;
   g_mutex_lock (&mElement->state_lock);
   mElement->started = FALSE;
-  g_cond_signal (&mElement->state_cond);
+  g_cond_broadcast (&mElement->state_cond);
```
MathieuDuponchelle, line 378: Why is this needed? Can there be multiple threads waiting on this condition variable?

gstcefsrc.cc
```diff
@@ -352,37 +396,77 @@ void BrowserClient::MakeBrowser(int arg)
...
+  g_mutex_unlock(&init_lock);
 
   browser->GetHost()->SetAudioMuted(true);
 
   mElement->browser = browser;
 
   g_mutex_lock (&mElement->state_lock);
   mElement->started = TRUE;
-  g_cond_signal (&mElement->state_cond);
+  g_cond_broadcast  (&mElement->state_cond);
```
MathieuDuponchelle, line 411: and here

gstcefsrc.cc
```diff
@@ -525,66 +654,111 @@ run_cef (GstCefSrc *src)
...
+    g_mutex_lock (&init_lock);
+    // Wait till a previous CEF is dismantled or completes initialization
+    while (cef_shutdown || cef_initializing)
       g_cond_wait (&init_cond, &init_lock);
+    if (cef_inited && !init_result) {
+      // BAIL OUT, CEF is not loaded.
+      result = GST_STATE_CHANGE_FAILURE;
+    } else if (!cef_inited) {
+      cef_initializing = TRUE;
+      g_cond_broadcast (&init_cond);
```
MathieuDuponchelle, line 695: That's not wrong, but I don't think anything is waiting for `cef_initializing = TRUE` is there?

gstcefsrc.cc
```diff
@@ -624,6 +795,12 @@ gst_cef_src_start(GstBaseSrc *base_src)
   return ret;
 }
 
+static void
+gst_close_browser(GstCefSrc *src)
```
MathieuDuponchelle, line 799: `gst_cef_close_browser` would be more inline with the rest of the names :)

gstcefsrc.cc
```diff
@@ -487,23 +587,52 @@ run_cef (GstCefSrc *src)
...
+#else
+  gchar* browser_subprocess_path = g_build_filename(base_path, "gstcefsubprocess", nullptr);
+#endif
   if (const gchar *custom_subprocess_path = g_getenv ("GST_CEF_SUBPROCESS_PATH")) {
     g_setenv ("CEF_SUBPROCESS_PATH", browser_subprocess_path, TRUE);
     g_free (browser_subprocess_path);
     browser_subprocess_path = g_strdup (custom_subprocess_path);
   }
 
+  GST_WARNING_OBJECT(src, "CEF subprocess: %s\n", browser_subprocess_path);
```
MathieuDuponchelle, line 607: Probably want `GST_DEBUG_OBJECT` at most here ?

gstcefsrc.cc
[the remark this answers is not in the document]
sdroege: C++ has various type complications, and you should really use `nullptr` instead of `NULL`. `NULL` is an integer 0 in C++ (so you might or might not get a pointer-sized 0 here with the varargs depending on a lot of stuff), while `nullptr` is a pointer

gstcefsrc.cc
```diff
@@ -525,66 +654,111 @@ run_cef (GstCefSrc *src)
     /* unblock start () */
     g_mutex_lock (&init_lock);
     cef_inited = TRUE;
-    g_cond_signal(&init_cond);
+    g_cond_broadcast (&init_cond);
     g_mutex_unlock (&init_lock);
 
-    goto done;
+    return NULL;
```
MathieuDuponchelle, line 660: nit: I like single exit points functions if you don't mind keeping it that way :)

gstcefsrc.cc
```diff
@@ -525,66 +654,111 @@ run_cef (GstCefSrc *src)
...
-
-done:
+  gst_cef_shutdown(nullptr);
+#endif
   return NULL;
 }
 
-void quit_message_loop (int arg)
+static GstStateChangeReturn
+gst_cef_handle_lifetime(GstElement *element, GstStateChange transition)
```
MathieuDuponchelle, line 678: nit: convention would be to name this `gst_cef_src_set_state` :)

gstcefsrc.cc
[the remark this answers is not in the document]
nirbheek: glib defines `NULL` as `nullptr` when C++11 is being used:

```c
#ifndef NULL
#  if G_CXX_STD_CHECK_VERSION (11)
#    define NULL (nullptr)
#  elif defined (G_CXX_STD_VERSION)
#    define NULL (0L)
#  else
#    define NULL ((void*) 0)
#  endif /* G_CXX_STD_CHECK_VERSION (11) */
#endif
```

Perhaps this doesn't work correctly on clang, or something else is defining `NULL` here.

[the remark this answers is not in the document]
sdroege: You can't rely on GLib's definition to be used. Every library has code like this, and until C++11 the correct definition of `NULL` was indeed `0L` (reason: `void *` does not implicitly convert to any other pointer type with C++ but a 0 integer does :clown_face: )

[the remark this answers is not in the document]
MathieuDuponchelle: oof OK, thanks for the explanation :)

gstcefloader.cc
```diff
@@ -0,0 +1,107 @@
...
+    return {};
+  }
+
+  std::ostringstream ss;
+  ss << parent_dir.get() << "/" << kFrameworkPath;
+  return ss.str();
+}
+} // namespace
+
+std::string gst_get_cef_framework_path(bool helper)
```
MathieuDuponchelle, line 73: `gst_cef_get_framework_path` :)

gstcefnsapplication.mm
```diff
@@ -0,0 +1,99 @@
...
+
+- (void)_swizzled_sendEvent:(NSEvent *)event {
+  CefScopedSendingEvent sendingEventScoper;
+  // syslog(LOG_DEBUG, "[NSEvent sendEvent]: %p", event);
+  // Calls NSApplication::sendEvent due to the swizzling.
+  [self _swizzled_sendEvent:event];
+}
+@end
+
+BOOL gst_is_main_thread()
```
MathieuDuponchelle, line 96: `gst_cef_is_main_thread` :)

MathieuDuponchelle:
I have no other remarks, this is ready to merge for me after those are addressed, that's great work @amyspark thanks for tackling that, and thank you for the sponsoring @reinismu !

gstcefsrc.cc
amyspark: Makes sense. I had this as a warning as I traced what data was making its way to libcef.

gstcefsrc.cc
amyspark: If multiple pipelines are created simultaneously with CEF down, only one must run the initializer. `!cef_inited` is enough to check for initialization completion, but in the case of thread-based initialization, I think a second check is needed to prevent multiple threads.

gstcefsrc.cc
amyspark: I don't really think so, as each browser belongs to a single pipeline/element, and all of them run in CEF's UI thread. (This was probably overkill from attempting to address @sdroege's concern about thread waiting.)

gstcefsrc.cc
amyspark: Interesting, that looks nice!

gstcefsrc.cc
MathieuDuponchelle: hm, this isn't a check though, this is simply notifying other threads that the status has moved to "Initializing" :)

In practice, the broadcast that will matter here is the one done after the status has been set to "initialized", either from this thread or from another one.

amyspark:
Working on those nits, I'm marking them as resolved as I go through them in VS Code. I'll push the amended commits once done 👍

gstcefsrc.cc
MathieuDuponchelle: ack, that's what I imagined, let's keep the `g_cond_signal` here then :)

amyspark:
Nits addressed and initialization variables converted (thanks @MathieuDuponchelle for the insight!), I'd need one more look before merging. @sdroege ?

gstcefsrc.cc
amyspark: I've removed this broadcast in favour of the one produced by the initialization routine.

nirbheek:
If you haven't already @amyspark, I would suggest testing it once last time on macOS before merging, because non-trivial changes were made to the code.

amyspark:
@nirbheek I did just before pushing, I found another race condition in the event pump ([active timers must always be cancelled if the pump is called, irrespective of the plugin's status](https://github.com/centricular/gstcefsrc/compare/4ca3437a739700a9c726248f165ada8169a02de2..d61bb9153f46718187830011b2f1fec6f2c9d267#diff-b511dd6555eaa8cc62167e0332cbbda8f626ad7900fa3a9b092292a6c07e1ac5L427-R440)). Both macOS and Windows launch, run and exit without any issues.

gstcefsrc.cc
```diff
@@ -21,14 +31,45 @@ GST_DEBUG_CATEGORY_STATIC (cef_console_debug);
...
 #define DEFAULT_SANDBOX FALSE
+#endif
+
+using CefStatus = enum : guint8 {
+  // CEF was either unloaded successfully or not yet loaded.
+  CEF_STATUS_NOT_LOADED = 0U,
+  // Blocks other elements from initializing CEF is it's already in progress.
+  CEF_STATUS_INITIALIZING = 1U,
+  // CEF's initialization process has completed successfully.
+  CEF_STATUS_INITIALIZED = 1U << 2U,
```
MathieuDuponchelle, line 47: I don't think the bit shifts are needed, unless you wanted to use those as flags but that was not exactly what I was suggesting at least :)

amyspark: That was indeed the purpose :) By turning them into flags, I could make masks to be used with the busy loops (just below this line)

gstcefsrc.cc
```diff
@@ -524,67 +656,109 @@ run_cef (GstCefSrc *src)
...
+#endif
+      }
+    }
+    g_mutex_unlock(&init_lock);
+    break;
+  }
+  case GST_STATE_CHANGE_READY_TO_NULL:
+  {
+    g_mutex_lock (&init_lock);
+    if (cef_status == CEF_STATUS_SHUTTING_DOWN) {
```
sdroege, line 731: Multiple elements could get into this `if` and then on Apple call `gst_cef_shutdown()`. Should that be avoided? Then you'd have to change `cef_status` before the `g_cond_wait()` to something else

gstcefsrc.cc
```diff
@@ -632,7 +809,11 @@ gst_cef_src_stop (GstBaseSrc *base_src)
   GST_INFO_OBJECT (src, "Stopping");
 
   if (src->browser) {
-    src->browser->GetHost()->CloseBrowser(true);
+#ifdef __APPLE__
+    dispatch_async_f(dispatch_get_main_queue(), src, (dispatch_function_t)&gst_cef_src_close_browser);
+#else
+    gst_cef_src_close_browser(src);
```
sdroege, line 815: Here and elsewhere, on non-Apple does this have to be called from the event loop thread or can it be called from anywhere?

gstcefnsapplication.h
```diff
@@ -0,0 +1,23 @@
...
+/// slice of work. The handling will take place in the main thread.
+/// The function does not block.
+CFRunLoopTimerRef gst_cef_domessagework(CFTimeInterval interval);
+/// Hook on different lifecycle events to trigger CefShutdown.
+/// These are:
+/// - NSApplicationWillTerminateNotification
+/// - NSEvent.NSEventMaskApplicationDefined as sent by gstmacos.mm
+/// The shutdown executes only once.
+void gst_cef_set_shutdown_observer();
+/// Execute [NSThread isMainThread] from a non-Objective-C context.
```
sdroege, line 22: Could also call `pthread_main_np()` from C but I guess it doesn't really matter

gstcefsrc.cc
MathieuDuponchelle: OK, carry on then :)

gstcefsrc.cc
MathieuDuponchelle: Yeah, shouldn't the implementation instead read:

```
      while (cef_status & CEF_STATUS_MASK_TRANSITIONING)
        g_cond_wait (&init_cond, &init_lock);
      /* At that point the element holds the lock and knows no status change is happening */
      if (browsers == 0 && cef_status != CEF_STATUS_NOT_LOADED) {
         /* This element is the one in charge of shutting down */
         cef_status = CEF_STATUS_SHUTTING_DOWN;
         /* Here quit message loop if needed, dispatch cef_shutdown otherwise */
         ...
         /* Now wait for shut down to be over */
         while (cef_status != CEF_STATUS_NOT_LOADED)
              g_cond_wait (&init_cond, &init_lock);
         /* Clean up */
         ...
      }
```

This also means that `BrowserClient::OnBeforeClose` should now only update the browsers count :)

gstcefsrc.cc
```diff
@@ -487,23 +590,52 @@ run_cef (GstCefSrc *src)
...
   g_free(browser_subprocess_path);
 
-  gchar *locales_dir_path = g_build_filename(base_path, "locales", NULL);
+#ifdef __APPLE__
+  const std::string framework_folder = []() {
+    std::string framework = gst_cef_get_framework_path(false);
+    const auto split = framework.find_last_of('/');
+    return framework.substr(0, split);
+  }();
+  GST_WARNING_OBJECT(src, "CEF framework_dir_path: %s\n", framework_folder.c_str());
```
MathieuDuponchelle, line 620: Sorry I also missed this, `GST_DEBUG` :)

gstcefsrc.cc
```diff
@@ -487,23 +590,52 @@ run_cef (GstCefSrc *src)
...
+  CefString(&settings.framework_dir_path).FromString(framework_folder);
+  const std::string main_bundle_folder = [&](){
+    uint32_t size = 0;
+    _NSGetExecutablePath(nullptr, &size);
+    std::vector<char> host(size);
+    _NSGetExecutablePath(host.data(), &size);
+    auto host2 = std::unique_ptr<char>(realpath(host.data(), nullptr));
+    GST_WARNING_OBJECT(src, "Main executable path: %s\n", host2.get());
```
MathieuDuponchelle, line 628: and here :)

reinismu:
Found one issue, but not sure if it is related to changes here or not.

Build works fine for me on MacOS, but if I try to package it together with Nix I run into issue
`ld: library not found for -lsandbox`

Here is the nix file https://gist.github.com/reinismu/31452372283a7b89715399bd7e763aad

I did some investigation and came to that clang that is installed by Nix only throws this.

Not working command
```
/nix/store/24ljvc5iwbs01svv9s8zvfcl5qs876kp-clang-wrapper-16.0.6/bin/clang++ -O3 -DNDEBUG -arch arm64 -mmacosx-version-min=10.15 -Wl,-search_paths_first -Wl,-headerpad_max_install_names -Wl,-search_paths_first -Wl,-ObjC -Wl,-pie -Wl,-dead_strip -L/nix/store/wmkr30nbg14vrg08xxvyggwgkixppa3g-gstreamer-1.22.9/lib -L/nix/store/pnafnhig3vz9h0zixvp2gxr5wcywq7kz-glib-2.78.4/lib -L/nix/store/yjzrvf8mppc3yj3qyixkjpvqar0w900k-gst-plugins-base-1.22.9/lib -lgstvideo-1.0 -lgstaudio-1.0 -lgstbase-1.0 -lgstreamer-1.0 -Wl,-rpath,/nix/store/wmkr30nbg14vrg08xxvyggwgkixppa3g-gstreamer-1.22.9/lib -lgobject-2.0 -lglib-2.0 CMakeFiles/gstcefsubprocess.dir/gstcefloader.cc.o CMakeFiles/gstcefsubprocess.dir/gstcefsubprocess.cc.o -o Release/gstcefsubprocess.app/Contents/MacOS/gstcefsubprocess  libcef_dll_wrapper/libcef_dll_wrapper.a -lpthread -framework Cocoa -framework AppKit -lsandbox /tmp/nix-build-gstcefsrc-0.1.0.drv-0/source/third_party/cef/cef_binary_122.1.13+gde5b724+chromium-122.0.6261.130_macosarm64/Release/cef_sandbox.a /nix/store/pnafnhig3vz9h0zixvp2gxr5wcywq7kz-glib-2.78.4/lib/libglib-2.0.dylib -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks
```

if I call it with `/Library/Developer/CommandLineTools/usr/bin/c++` and the same arguments it works

I tried both V15 and V16 Clang and they had the same issue

gstcefnsapplication.h
amyspark: It's perfectly OK, one less function in the Objective-C bridge 😁

gstcefsrc.cc
amyspark: On checking in my Mac (it's the only one where this can be life-or-death of CEF) it can be called from anywhere, CEF itself posts the event to the UI thread. I'll clean this up.

gstcefsrc.cc
amyspark: Just applied this, with one nit: the check must be against `CEF_STATUS_INITIALIZED`, otherwise CEF asserts because there's not a global context active.

sdroege: You forgot to push this or is github confused?

amyspark: Let me see, I think it got pushed in the wrong repo.

gstcefsrc.cc
```diff
@@ -524,67 +653,109 @@ run_cef (GstCefSrc *src)
...
+      g_cond_wait (&init_cond, &init_lock);
+    /* At that point the element holds the lock and knows no status change is happening */
+    if (browsers == 0 && cef_status == CEF_STATUS_INITIALIZED) {
+      /* This element is the one in charge of shutting down */
+      cef_status = CEF_STATUS_SHUTTING_DOWN;
+#ifdef __APPLE__
+      /* in the main thread as per Cocoa */
+      dispatch_async_f(dispatch_get_main_queue(), nullptr, (dispatch_function_t)&gst_cef_shutdown);
+#endif
+      // the UI thread handles it through the message loop return,
```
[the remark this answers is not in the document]
amyspark: CEF requires that `CefQuitMessageLoop` must be issued from inside the UI thread; calling it from outside of the loop will trigger a thread validation assertion.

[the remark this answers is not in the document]
MathieuDuponchelle: Actually, you could also move the`browsers` decrementation here and not touch `init_lock` or `init_cond` at all from `OnBeforeClose`, I think that will make things even easier to reason about

sdroege (approved):

gstcefsrc.cc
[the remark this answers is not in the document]
MathieuDuponchelle: my bad, didn't see your last comment, the code is perfect as is then :)

MathieuDuponchelle:
Let's gooo! Thanks again everyone, this is so nice to finally see this solved :)

amyspark:
@reinis No, that commit was related to an issue one of my colleagues found, CMake runs all steps linearly, and the last `install(DIRECTORY... DESTINATION .)` was copying over `gstcefsubprocess` without the executable bit set.

Can you send me the recipe you're using? Either here or through email (amy at centricular).

sidsethupathi:
> I assume I will need to make it as an MacOS app as well for Cef to work properly

Has anyone done this with a Rust development environment? I currently am using the threaded callback function from the [`gstreamer-rs` examples](https://github.com/sdroege/gstreamer-rs/blob/192844c17363aedceffb32d2b726a5642bb8183e/examples/src/examples-common.rs) and was starting to go down the path of adding this: https://github.com/mrmekon/fruitbasket, but wondering if anyone already has a solution for it.

After following the instructions in the README, `gst-launch-1.0` works fine (eg. `gst-launch-1.0 cefsrc ! autovideosink`) but when running a Rust program that starts a pipeline with cefsrc, it gets terminated by SIGTRAP, which is what I think I was seeing with `gst-launch-1.0` before doing the mac app workaround.

PR #86, opened 2024-07-04
Fix deadlock on macOS when running GStreamer in the main thread

amyspark:
Hi,

This addresses the issue in https://github.com/centricular/gstcefsrc/pull/83#issuecomment-2208943286 where CEF would deadlock on launch on macOS. I traced it to an incorrect assumption in my code, that GStreamer would always run in a separate thread; in the supplied app, GStreamer is handled from the Cocoa event loop.

While this was already addressed for the CEF startup itself, it didn't account for GStreamer's code flow before CEF initialization (GstElement's start handler runs before the call to `run_cef`) and on shutdown (there was no check for main thread, so the deadlock would've occurred on the call to `dispatch_async_f`).

Let me know if this is OK; I tested it locally with `gst-launch-1.0` and observed no issues.

cc @reinismu , can you check with your Nix recipe?

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -677,7 +677,7 @@ done:
 static GstStateChangeReturn
 gst_cef_src_change_state(GstElement *element, GstStateChange transition)
 {
-  auto result = GST_CALL_PARENT_WITH_DEFAULT (GST_ELEMENT_CLASS , change_state, (element, transition), GST_STATE_CHANGE_FAILURE);
+  GstStateChangeReturn result = GST_STATE_CHANGE_SUCCESS;
 
   switch(transition)
   {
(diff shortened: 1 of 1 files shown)


gstcefsrc.cc
```diff
@@ -1,3 +1,4 @@
+#include "gst/gstelement.h"
```
sdroege, line 1: Not needed

amyspark:
I found another one, this one comes from existing code. `CefPostTask` does not seem to if it's running in the main thread, so it'll post the task asynchronously. Then the semaphore will wait endlessly. I've added the same checking code as in `src_start` and `src_stop`.

With that I get the app running, it seems to be waiting for an outside connection -- I take that's expected @reinismu .

gstcefsrc.cc
```diff
@@ -778,13 +789,22 @@ gst_cef_src_start(GstBaseSrc *base_src)
   GST_OBJECT_UNLOCK (src);
 
   browserClient = new BrowserClient(renderHandler, audioHandler, requestHandler, displayHandler, src);
-  CefPostTask(TID_UI, base::BindOnce(&BrowserClient::MakeBrowser, browserClient.get(), 0));
+#ifdef __APPLE__
+  if (pthread_main_np()) {
```
sdroege, line 793: A similar change should be needed in `stop()` then? You'd close the browser there and wait for it to be closed, but if that is called from the main thread this would never happen because the closing would happen from that very thread's event loop?

sdroege:
`set_caps()` and `set_property()` are calling browser APIs but those would be called from arbitrary threads. Is that OK?

gstcefsrc.cc
amyspark: Yeah, that's a correct insight. Calls to CEF can be posted from either the main thread or the creator's thread, but the wait must happen only if the threads are different.

amyspark:
> set_caps() and set_property() are calling browser APIs but those would be called from arbitrary threads. Is that OK?

Yes, that's OK. They get processed by separate threads within CEF.

sdroege:
> Yes, that's OK. They get processed by separate threads within CEF.

This API is very confusing :)

gstcefsrc.cc
```diff
@@ -778,13 +789,22 @@ gst_cef_src_start(GstBaseSrc *base_src)
   GST_OBJECT_UNLOCK (src);
```
sdroege, line 789: The `g_cond_wait()` 10 lines above would also wait forever if we end up here on macOS on the main thread or not?

amyspark: Not after this MR. If GStreamer is managed from the main thread, this function gets called by line 762 above, by which time either CEF has initialized (skipping the wait altogether), or it has failed (and this function will not be called at all). The wait is still necessary if the app handles CEF in a different event loop from the GStreamer pipeline's, like with `gst-launch-1.0`.

PR #87, opened 2024-07-05
Fix GLib critical warning about default enum value not found

vivia:
LOGSEVERITY_DEFAULT is defined as a different enum value as LOGSEVERITY_INFO, which it corresponds to, and set as equal later on internally. This cannot work when looking it up in the GEnumValue array. Easiest way around it is to manually set it to INFO.


MathieuDuponchelle:
cc @amyspark

amyspark (approved):

PR #88, opened 2024-08-22
Add javascript signalling

aiden-jeffrey:
Based on [this sample](https://bitbucket.org/chromiumembedded/cef-project/src/master/examples/message_router), this adds a way to signal in javascript (from the webpage) a couple of things:
- "ready" - page is ready for rendering, so enable cefsrc to go to a PAUSED/PLAYING state
- "eos" - page doesn't have any more content, so push an EOS event onto cefsrc

This is useful for pages that have non-trivial load times and may run out of content (webgl pages, pages with videos, etc).

There is a sample in the html/ directory that shows how to send the signals, e.g.:
```
window.gstSendMsg({
          request: "ready",
          onSuccess: function(response) { ... }
          },
          onFailure: function(error_code, error_message) { ... }
        });
      }
```
To test, run a webserver in the html directory, say:
```
python -m http.server 9000
```
Then just specify the flag in a run pointing at that page:
```
gst-launch-1.0 -e cefsrc url="http://localhost:9000/ready_test.html" listen-for-js-signals=true ! video/x-raw, width=1920, height=1080, framerate=60/1 ! cefdemux name=d d.video ! queue max-size-bytes=0 max-size-buffers=0 max-size-time=3000000000 ! videoconvert ! xvimagesink
```

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -1288,7 +1288,10 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
           DEFAULT_SANDBOX, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
   g_object_class_install_property (gobject_class, PROP_LISTEN_FOR_JS_SIGNAL,
     g_param_spec_boolean ("listen-for-js-signals", "listen-for-js-signals",
-          "Listen and respond to signals sent from javascript: window.gstSendMsg({request: \"ready|eos\", ...})",
+          "Listen and respond to signals sent from javascript: "
+          "window.gstSendMsg({request: \"ready|eos\", ...}) - "
+          "see [README](https://github.com/centricular/gstcefsrc?tab=readme-ov-file#javascript-signals) "
+          "for more detail",
           DEFAULT_LISTEN_FOR_JS_SIGNALS, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
 
   g_object_class_install_property (gobject_class, PROP_JS_FLAGS,
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -23,6 +23,16 @@
 #include "gstcefnsapplication.h"
 #endif
 
+#define GST_ELEMENT_PROGRESS(el, type, code, text)      \
+G_STMT_START {                                          \
+  gchar *__txt = _gst_element_error_printf text;        \
+  gst_element_post_message (GST_ELEMENT_CAST (el),      \
+      gst_message_new_progress (GST_OBJECT_CAST (el),   \
+          GST_PROGRESS_TYPE_ ##type, code, __txt));     \
+  g_free (__txt);                                       \
+} G_STMT_END
+
+
 GST_DEBUG_CATEGORY_STATIC (cef_src_debug);
 #define GST_CAT_DEFAULT cef_src_debug
 
@@ -900,7 +910,11 @@ gst_cef_src_start(GstBaseSrc *base_src)
 {
   gboolean ret = FALSE;
   GstCefSrc *src = GST_CEF_SRC (base_src);
+
+  GST_ELEMENT_PROGRESS(src, START, "open", ("Creating CEF browser client"));
+
   CefRefPtr<BrowserClient> browserClient = new BrowserClient(src);
+  gulong browser_id = browsers;
 
   /* Make sure CEF is initialized before posting a task */
   g_mutex_lock (&init_lock);
@@ -908,12 +922,17 @@ gst_cef_src_start(GstBaseSrc *base_src)
     g_cond_wait (&init_cond, &init_lock);
   g_mutex_unlock (&init_lock);
 
-  if (cef_status == CEF_STATUS_FAILURE)
+  if (cef_status == CEF_STATUS_FAILURE) {
+    GST_ELEMENT_PROGRESS(src, ERROR, "open", ("CEF in failed state"));
     goto done;
+  }
 
   GST_OBJECT_LOCK (src);
   src->n_frames = 0;
   GST_OBJECT_UNLOCK (src);
+
+  GST_ELEMENT_PROGRESS(src, CONTINUE, "open", ("Creating CEF browser (#%lu)...", browser_id));
+
 #ifdef __APPLE__
   if (pthread_main_np()) {
     /* in the main thread as per Cocoa */
@@ -923,6 +942,8 @@ gst_cef_src_start(GstBaseSrc *base_src)
     CefPostTask(TID_UI, base::BindOnce(&BrowserClient::MakeBrowser, browserClient.get(), 0));
 
     /* And wait for this src's browser to have been created */
+    GST_ELEMENT_PROGRESS(src, CONTINUE, "open", ("Waiting for CEF browser initialization..."));
+
     g_mutex_lock(&src->state_lock);
     while (!CefSrcStateIsOpen(src->state))
       g_cond_wait (&src->state_cond, &src->state_lock);
@@ -930,6 +951,7 @@ gst_cef_src_start(GstBaseSrc *base_src)
 #ifdef __APPLE__
   }
 #endif
+
 
   if (src->listen_for_js_signals) {
     g_mutex_lock (&src->state_lock);
@@ -939,6 +961,18 @@ gst_cef_src_start(GstBaseSrc *base_src)
   }
 
   ret = src->browser != NULL;
+
+  if (ret) {
+    GST_ELEMENT_PROGRESS(
+      src, COMPLETE, "open",
+      ("CEF browser created (#%lu)", browser_id)
+    );
+  } else {
+    GST_ELEMENT_PROGRESS(
+      src, ERROR, "open",
+      ("CEF browser failed to create (#%lu)", browser_id)
+    );
+  }
 
 done:
   return ret;
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -166,8 +166,8 @@ class RenderHandler : public CefRenderHandler
 {
   public:
 
-    RenderHandler(GstCefSrc *element) :
-        element (element)
+    RenderHandler(GstCefSrc *src) :
+        src (src)
     {
     }
 
@@ -177,32 +177,32 @@ class RenderHandler : public CefRenderHandler
 
     void GetViewRect(CefRefPtr<CefBrowser> browser, CefRect &rect) override
     {
-	  GST_LOG_OBJECT(element, "getting view rect");
-      GST_OBJECT_LOCK (element);
-      rect = CefRect(0, 0, element->vinfo.width ? element->vinfo.width : DEFAULT_WIDTH, element->vinfo.height ? element->vinfo.height : DEFAULT_HEIGHT);
-      GST_OBJECT_UNLOCK (element);
+	  GST_LOG_OBJECT(src, "getting view rect");
+      GST_OBJECT_LOCK (src);
+      rect = CefRect(0, 0, src->vinfo.width ? src->vinfo.width : DEFAULT_WIDTH, src->vinfo.height ? src->vinfo.height : DEFAULT_HEIGHT);
+      GST_OBJECT_UNLOCK (src);
     }
 
     void OnPaint(CefRefPtr<CefBrowser> browser, PaintElementType type, const RectList &dirtyRects, const void * buffer, int w, int h) override
     {
       GstBuffer *new_buffer;
 
-      GST_LOG_OBJECT (element, "painting, width / height: %d %d", w, h);
+      GST_LOG_OBJECT (src, "painting, width / height: %d %d", w, h);
 
-      new_buffer = gst_buffer_new_allocate (NULL, element->vinfo.width * element->vinfo.height * 4, NULL);
+      new_buffer = gst_buffer_new_allocate (NULL, src->vinfo.width * src->vinfo.height * 4, NULL);
       gst_buffer_fill (new_buffer, 0, buffer, w * h * 4);
 
-      GST_OBJECT_LOCK (element);
-      gst_buffer_replace (&(element->current_buffer), new_buffer);
+      GST_OBJECT_LOCK (src);
+      gst_buffer_replace (&(src->current_buffer), new_buffer);
       gst_buffer_unref (new_buffer);
-      GST_OBJECT_UNLOCK (element);
+      GST_OBJECT_UNLOCK (src);
 
-      GST_LOG_OBJECT (element, "done painting");
+      GST_LOG_OBJECT (src, "done painting");
     }
 
   private:
 
-    GstCefSrc *element;
+    GstCefSrc *src;
 
     IMPLEMENT_REFCOUNTING(RenderHandler);
 };
@@ -211,8 +211,8 @@ class AudioHandler : public CefAudioHandler
 {
   public:
 
-    AudioHandler(GstCefSrc *element) :
-        mElement (element)
+    AudioHandler(GstCefSrc *src) :
+        src (src)
     {
     }
 
@@ -233,9 +233,9 @@ class AudioHandler : public CefAudioHandler
     mRate = params.sample_rate;
     mChannels = channels;
 
-    GST_OBJECT_LOCK (mElement);
-    mElement->audio_events = g_list_append (mElement->audio_events, event);
-    GST_OBJECT_UNLOCK (mElement);
+    GST_OBJECT_LOCK (src);
+    src->audio_events = g_list_append (src->audio_events, event);
+    GST_OBJECT_UNLOCK (src);
   }
 
   void OnAudioStreamPacket(CefRefPtr<CefBrowser> browser,
@@ -247,7 +247,7 @@ class AudioHandler : public CefAudioHandler
     GstMapInfo info;
     gint i, j;
 
-    GST_LOG_OBJECT (mElement, "Handling audio stream packet with %d frames", frames);
+    GST_LOG_OBJECT (src, "Handling audio stream packet with %d frames", frames);
 
     buf = gst_buffer_new_allocate (NULL, mChannels * frames * 4, NULL);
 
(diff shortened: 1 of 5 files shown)


gstcefsrc.h
```diff
@@ -69,7 +69,8 @@ class App : public CefApp, public CefBrowserProcessHandler {
   void OnScheduleMessagePumpWork(int64_t delay_ms) override;
 #endif
 private:
-  IMPLEMENT_REFCOUNTING(App);
+  IMPLEMENT_REFCOUNTING(BrowserApp);
+  DISALLOW_COPY_AND_ASSIGN(BrowserApp);
```
MathieuDuponchelle, line 74: I'm curious what that is for?

MathieuDuponchelle:
Can you expand on the commit message for "message: Add message routers and handlers ", explaining why this is actually needed? The code looks fine but I can't quite understand the intent behind it :)

gstcefsrc.h
aiden-jeffrey: To be honest my thinking didn't go beyond "the sample has it, so maybe I should add it" - https://github.com/chromiumembedded/cef-project/blob/aaccb78661c390a5ce6650df8db54ea5424d431c/examples/message_router/app_browser_impl.cc#L53

gstcefsrc.cc
```diff
@@ -133,31 +142,72 @@ gchar* get_plugin_base_path () {
...
-    const std::string& url = frame->GetURL();
-    if (url.find(startup_url_) != 0)
-      return false;
+               CefRefPtr<Callback> callback) override
+  {
+    if (!src) return false;
+
+    // TODO: do we want to make the incoming payload json??
+    bool success = false;
+    std::ostringstream error_msg;
```
MathieuDuponchelle, line 166: shouldn't this be moved to the `else {}` ?

MathieuDuponchelle:
I know this wasn't the case before, but I think you will want to also add the posting of progress messages to `gst_cef_src_start()` (refer to eg rtspsrc for an example).

The start method was already blocking while waiting for cef initialization, which isn't great without posting progress messages, but if we  now also (optionally) wait for an arbitrary amount of time for the web page to reach a certain state it becomes even more critical to post these :)

You should notice the difference in behavior with gst-launch after the messages are added.

gstcefsrc.h
MathieuDuponchelle: OK, let's not add it here then I think :)

gstcefsrc.cc
```diff
@@ -1111,6 +1257,10 @@ gst_cef_src_class_init (GstCefSrcClass * klass)
     g_param_spec_boolean ("sandbox", "sandbox",
           "Toggle chromium sandboxing capabilities",
           DEFAULT_SANDBOX, (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | GST_PARAM_MUTABLE_READY)));
+  g_object_class_install_property (gobject_class, PROP_LISTEN_FOR_JS_SIGNAL,
```
MathieuDuponchelle, line 1260: Let's add a docstring here, and link to a section in the README explaining in detail the expected protocol between the element and the web page :)

MathieuDuponchelle:
Sorry for not reviewing earlier ! This looks very useful to me, though admittedly not entirely generic, I'm OK with merging once my comments are addressed :)

aiden-jeffrey:
> though admittedly not entirely generic

I toyed with allowing the user to send all gstreamer events from javascript, but then I'm not sure there is an event that correlates to the "ready" signal I add here. If you can think of a way of making this more flexible / generic, then I'd be happy to improve the API.

MathieuDuponchelle:
> If you can think of a way of making this more flexible / generic, then I'd be happy to improve the API.

No, I think that's fine for now :)

aiden-jeffrey:
> I know this wasn't the case before, but I think you will want to also add the posting of progress messages to `gst_cef_src_start()` (refer to eg rtspsrc for an example).

Is there a reason `cef_status` is file level static and not in the `GstCefSrc` struct?

aiden-jeffrey:
> Oh hrm wait, this question makes me realize you're probably piggy-backing the wrong mechanism :)

Yeah.

MathieuDuponchelle:
Here you want a separate, per source additional mechanism to wait for readiness, you can probably mirror the existing one

aiden-jeffrey:
Should I use the `state_lock` and `state_cond` for the wait mechanics?

MathieuDuponchelle:
Yes, that seems appropriate

MathieuDuponchelle:
Can you squash the following commits together:

> [signals: Add support for starting and stopping cefsrc from js](https://github.com/centricular/gstcefsrc/pull/88/commits/84e22caae447e600aaa3c482fe0e4a644dc385be)

> [signals: Move wait-ready mutex into src struct](https://github.com/centricular/gstcefsrc/pull/88/commits/bb2b114cd3b0667b1839a4414d9d395e8a43b2d1)

?

Re progress messages, that is not what I meant, here is an example:

https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-good/gst/rtsp/gstrtspsrc.c?ref_type=heads#L6319-6375

aiden-jeffrey:
> Re progress messages, that is not what I meant

So replace where I've added GST_DEBUG_OBJECT calls with your GST_ELEMENT_PROGRESS macro?

gstcefsrc.cc
```diff
@@ -901,19 +911,25 @@ gst_cef_src_start(GstBaseSrc *base_src)
...
+  if (cef_status == CEF_STATUS_FAILURE) {
+    GST_ELEMENT_PROGRESS(src, ERROR, "open", ("CEF in failed state"));
     goto done;
+  }
 
   GST_OBJECT_LOCK (src);
   src->n_frames = 0;
   GST_OBJECT_UNLOCK (src);
+
+  GST_ELEMENT_PROGRESS(src, START, "open", ("Creating cef browser (#%lu)...", browser_id));
```
MathieuDuponchelle, line 931: move this to before the wait for cef initialization

gstcefsrc.cc
```diff
@@ -931,11 +949,19 @@ gst_cef_src_start(GstBaseSrc *base_src)
   }
 #endif
 
+  GST_ELEMENT_PROGRESS(src, COMPLETE, "open", ("CEF browser created (#%lu)", browser_id));
```
MathieuDuponchelle, line 952: Move this to the end of the start implementation

gstcefsrc.cc
```diff
@@ -931,11 +949,19 @@ gst_cef_src_start(GstBaseSrc *base_src)
...
+
   if (src->listen_for_js_signals) {
+    GST_ELEMENT_PROGRESS(
```
MathieuDuponchelle, line 955: Remove this, we don't need to complicate things :)

gstcefsrc.cc
```diff
@@ -931,11 +949,19 @@ gst_cef_src_start(GstBaseSrc *base_src)
...
+      src, CONTINUE, "wait",
+      ("Waiting for window.gstSendMessage({\"msg\": \"ready\", ...}) from cef)")
+    );
     g_mutex_lock (&src->state_lock);
     while (src->state == CEF_SRC_WAITING_FOR_READY)
       g_cond_wait (&src->state_cond, &src->state_lock);
     g_mutex_unlock (&src->state_lock);
+
+    GST_ELEMENT_PROGRESS(src, COMPLETE, "wait", ("Ready signal received from CEF"));
```
MathieuDuponchelle, line 964: And remove this :)

MathieuDuponchelle:
Nearly there :)

MathieuDuponchelle:
lgtm now :)

aiden-jeffrey:
> lgtm now :)

Thanks for the review.

MathieuDuponchelle:
> Thanks for the review.

And thanks for the patch :)

SteveMcFarlin:
@aiden-jeffrey I was working on something similar a while back. Basically to allow communication between JS and the hosting app. In my case this would have been a Rust service. My branch is [here](https://github.com/SteveMcFarlin/gstcefsrc/tree/cef-ipc). It is probably overkill for this upstream project.

aiden-jeffrey:
> @aiden-jeffrey I was working on something similar a while back. Basically to allow communication between JS and the hosting app.

This looks very comprehensive. Our app is also a rust app.

SteveMcFarlin:
@aiden-jeffrey - it is if you want IPC between the browser and the client application. Essentially bidirectional comm over: 

JS - Render Process - Browser Process - GST Signal/Action - App.

I was writing the Rust app as a Node plugin via NAPI-RS. I thought it would be 'cool' to have this as a full communication channel. However, it is probably best to just use a websocket between the Browser JS and the hosting app and call it a day if you just need communication. In my case I needed to inject custom JS code into the window object. Similar to what Puppeteer does before.

SteveMcFarlin:
btw. Thanks for the patch.

aiden-jeffrey:
I'm getting a seg fault when starting another pipeline with a cefsrc in after the first cefsrc has emmited the eos due to the javascript `signal`. It segfaults somewhere in CefInitialize, but i can't work out where because I can't get gstreamer to pick up the plugin when built with `Debug` symbols.

MathieuDuponchelle:
ouch, I don't have an idea right now no

aiden-jeffrey:
> can you post a test case?

I don't have a gst-launch command that allows for this behaviour. Let me see if I can make a simplified version of our application to replicate. 

But the basic idea is:
- create two pipelines
- start one
- wait for the eos triggered via the js signal
- watch the cef_status go to 0
- start second one 
- see CefInitialize call segfault

MathieuDuponchelle:
OK, yes please make a reproducer application so I can take a look on my end, otherwise I'm afraid we'll have to revert this until we come with a fix for the issue :)

aiden-jeffrey:
@MathieuDuponchelle - my colleague set up a repro example last night - I think this is a bug that pre-existed this signals work. See https://github.com/aiden-jeffrey/test-cefsrc for the sample. There's two in there. I managed to repro the segfault on the sha prior to the signals work (002f660) using the `sleep` binary.

aiden-jeffrey:
It's possible I need to do more cleanup on the rust side. I'm not super confident in rust.

aiden-jeffrey:
@MathieuDuponchelle - I think it might be a regression. `5643c02bda8a67f9397aa15d4905a52c00d6cc16` doesn't have this crash.

I performed a rebase between 5643c02 and 002f660, I had to skip some commits, this was the result of my bisect:
```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
189f7bcca308dd0a2f93e8090ac3b153cd542bce
bf0c9420aac6d58446e5638b7803c8e0ef21e034
4e79b292a6eff3f1179b24329fc263bb84e979c5
bc511024498892cb271da756833c8c4fb674dbc2
6e469dc064cef5d78431285784fc20b0f2108a15
779761f653a7d6d95344699aa534723c95ab7a4e
1e4f194c5546bb1fa70e139d28277e9cfcc6f1cd
fb9be4f9e5a88c7ca3880ce0edce7dd7ec51de65
4b8a7447e154ef16327f4bcc3c38a01ef6c31f9e
8383a4c3a8012d9919aa7e1000807c2671b06fef
187b280ddadf388d29b83d57331d83602a193707
bfa3de0fdbcbc22d26e29076595c1ff7742e3b64
6f1823cf71234c7fd19261f53bb6754cadc7e380
We cannot bisect more!
```

MathieuDuponchelle:
@aiden-jeffrey yeah I was wondering if this was a regression from https://github.com/centricular/gstcefsrc/pull/83 , it looks likely that it is related. Why did you need to skip these commits? cc@amyspark

MathieuDuponchelle:
Thanks for the reproducer by the way

aiden-jeffrey:
> Why did you need to skip these commits?

I got some unrelated errors trying to run the `sleep` binary.

For half of these (for example 189f7bcca308dd0a2f93e8090ac3b153cd542bce) I got:
```
$ cargo run --bin sleep
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/sleep`

(gst-plugin-scanner:1632010): GLib-GObject-CRITICAL **: 13:33:35.045: g_param_spec_enum: assertion 'g_enum_get_value (enum_class, default_value) != NULL' failed

(gst-plugin-scanner:1632010): GLib-GObject-CRITICAL **: 13:33:35.045: validate_pspec_to_install: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(gst-plugin-scanner:1632010): GLib-GObject-CRITICAL **: 13:33:35.045: g_param_spec_ref_sink: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(gst-plugin-scanner:1632010): GLib-GObject-CRITICAL **: 13:33:35.045: g_param_spec_unref: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(sleep:1632002): GLib-GObject-CRITICAL **: 13:33:35.098: g_param_spec_enum: assertion 'g_enum_get_value (enum_class, default_value) != NULL' failed

(sleep:1632002): GLib-GObject-CRITICAL **: 13:33:35.098: validate_pspec_to_install: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(sleep:1632002): GLib-GObject-CRITICAL **: 13:33:35.098: g_param_spec_ref_sink: assertion 'G_IS_PARAM_SPEC (pspec)' failed

(sleep:1632002): GLib-GObject-CRITICAL **: 13:33:35.098: g_param_spec_unref: assertion 'G_IS_PARAM_SPEC (pspec)' failed
```

- for a few (e.g. bfa3de0fdbcbc22d26e29076595c1ff7742e3b64) I got:
```
GStreamer-WARNING **: 13:36:37.263: Failed to load plugin '/home/aiden/-/gstcefsrc/build/Release/libgstcef.so': /home/aiden/-/gstcefsrc/build/Release/libgstcef.so: undefined symbol: cef_string_multimap_free
```

MathieuDuponchelle:
I see, annoying that the history doesn't build cleanly

aiden-jeffrey:
> yeah I was wondering if this was a regression from #83 ,

I tried reverting this merge commit and the issue still persists.

aiden-jeffrey:
> I tried reverting this merge commit and the issue still persists.

See https://github.com/aiden-jeffrey/gstcefsrc/tree/aiden/try-revert.

MathieuDuponchelle:
Hm, there was no merge commit, but 14 commits merged fast forward, I'm currently building the branch right before the first one of those

MathieuDuponchelle:
That's #86 , not #83 :)

aiden-jeffrey:
Just spotted my mistake! Doh.

MathieuDuponchelle:
yeah, some commit in #83 introduced the issue, @amyspark  do you mind taking a look? `cargo run --bin sleep` in https://github.com/aiden-jeffrey/test-cefsrc segfaults with master, completes successfully with d33494c6a26d20576e3bcf54a15ca5c5259601f4 and prior :)

aiden-jeffrey:
Pushed a small change (removing log-severity) to that test repo. Now my bisect completes and identifies `bc511024498892cb271da756833c8c4fb674dbc2` as the first bad commit.

aiden-jeffrey:
I think the issue here is `CefShutdown` is getting called, which according to the header should only be called on application exit. I've removed that, but now can't get the `MakeBrowser` `CefTask` to run in the second pipeline.

aiden-jeffrey:
Ok. Multiple pipelines just aren't really supported the way this element is currently written. If you run two cefsrc elements at the same time (without calling CefShutdown), Initialize never gets acalled, which means that the second cefsrc just has the properties of the first.

MathieuDuponchelle:
@aiden-jeffrey CefInitialize being called a single time is not a problem, but yes the javascript flags property is only used for the first source, which isn't great, other properties don't affect `CefInitialize` afaict

MathieuDuponchelle:
Ah, `sandbox` too, perhaps others, anyway see https://github.com/centricular/gstcefsrc/pull/90 , CefInitialize simply can't be called more than once in a given process

SteveMcFarlin:
@aiden-jeffrey Nice this got merged. Just saw the lightning talk at this years GStreamer conference. I was doing something similar at StreamYard prior to the company being sold.

aiden-jeffrey:
> @aiden-jeffrey Nice this got merged. Just saw the lightning talk at this years GStreamer conference. I was doing something similar at StreamYard prior to the company being sold.

@SteveMcFarlin - I ended up feeling like my patch was a bit inflexible after seeing your branch. I've been wondering if it's doing something more open ended like your work, that would support other use cases too.

SteveMcFarlin:
> @SteveMcFarlin - I ended up feeling like my patch was a bit inflexible after seeing your branch. I've been wondering if it's worth doing something more open ended like your work, that would support other use cases too.

@aiden-jeffrey - The work I did is probably overkill for something simple and open ended. I wanted to be able to get all console logs from the browser (In case of unhandled exceptions), any internal CEF errors, and be able to attach arbitrary functions to the global `window` object. It's been a while since I looked at my code, but feel free to use it as a reference. A rough sketch of what I was doing is [here](https://drive.google.com/file/d/1Z4W32WANuYjeLCsW862ha9uUC7Q3Kyfs/view?usp=sharing). I was not going to use CefQuery. 'Simple' two way IPC. 

If I were going to write something simple but yet flexible I would probably just expose the IPC between the Render and Browser processes. So basically on the Render side you would attach two functions to the browser's window object. Say `window.CefHandleEvent` and `window.CefSendEvent`. On the CefSrc side I would expose two signals. Say `on-cef-event`, and an action signal `send-cef-event`. I would probably force the use of strings. This would be the most flexible given the vast majority of use cases are going to be passing JSON blobs.

Anyway. My early morning thoughts on this. I had intended to do something generic like this for upstream, but just never got around to it.

SteveMcFarlin:
Btw. In my code I was not going to use this pattern:

`window.gstSendMsg({
          request: "ready",
          onSuccess: function(response) { ... }
          },
          onFailure: function(error_code, error_message) { ... }
        });
      }
`
If I recall correclty this is the `CefQuery` way of doing this. Take a look at [renderprocesshandler.cc](https://github.com/SteveMcFarlin/gstcefsrc/blob/cef-ipc/renderprocesshandler.cc). I am attaching functions there. The is also `RenderProcessHandler::OnProcessMessageReceived`. That is where browser process message are handled. Then in [browserclient.cc](`https://github.com/SteveMcFarlin/gstcefsrc/blob/cef-ipc/browserclient.cc`) there is `BrowserClient::OnProcessMessageReceived` where message from the render process are handled... e.g. JS functions called on the window object.

I don't recall how this is hooked up, but I know it's just the simple IPC mechanism provided by Chromium (Maybe CEF?). I have a modified [message_router](https://github.com/SteveMcFarlin/cef-project/tree/main/examples/message_router) example here that I used to test this out with. Basically I found the CefQuery pattern very cumbersome, and did not want to use it. I also sorta followed [this](https://bitbucket.org/chromiumembedded/cef/wiki/GeneralUsage#markdown-header-generic-message-router) doc.

SteveMcFarlin:
One final thought here. The simple idea is to use the IPC between the Browser and Render processes to pass messages. On the Render side use the V8 engine to add some function(s) to the window object. On the Browser process side (CefSrc) add some signals. The only real thing there is creating some 'header' to the IPC packets for routing. Feel free to ask me anything.  You can PM me [here](https://discourse.gstreamer.org/u/stevemcfarlin/summary)

MathieuDuponchelle:
Thanks a lot for taking the time to share your ideas @SteveMcFarlin :)

PR #90, opened 2024-10-04
cefsrc: fix CefInitialize crashing the second time

MathieuDuponchelle:
Calling CefInitialize multiple times during a process lifetime is not supported:

* https://magpcss.org/ceforum/viewtopic.php?f=6&t=16430
* https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=466
* https://stackoverflow.com/questions/73172139/cef-can-only-be-initialized-once-per-process-this-is-a-limitation-of-the-underl

Prior to #83 we used to have a ShutdownEnforcer class, but as it was causing issues on OSX we will simply no longer shut down CEF at all.

We can revisit if needed, but this is preferable to crashing.

See https://github.com/centricular/gstcefsrc/pull/88#issuecomment-2388711778 and below

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -62,19 +62,12 @@ using CefStatus = enum : guint8 {
   CEF_STATUS_INITIALIZED = 1U << 2U,
   // No CEF elements will be allowed to complete initialization.
   CEF_STATUS_FAILURE = 1U << 3U,
-  // CEF has been marked for shutdown, stopping any leftover events from being
-  // processed. Any elements that need it must wait till this flag and
-  // cef_initializing are cleared.
-  CEF_STATUS_SHUTTING_DOWN = 1U << 4U,
 };
+
 static CefStatus cef_status = CEF_STATUS_NOT_LOADED;
 static const guint8 CEF_STATUS_MASK_INITIALIZED = CEF_STATUS_FAILURE | CEF_STATUS_INITIALIZED;
-static const guint8 CEF_STATUS_MASK_TRANSITIONING =
-  CEF_STATUS_SHUTTING_DOWN | CEF_STATUS_INITIALIZING;
+static const guint8 CEF_STATUS_MASK_TRANSITIONING = CEF_STATUS_INITIALIZING;
 
-// Number of running CEF instances. Setting this to 0 must be accompanied
-// with cef_shutdown to prevent leaks on application exit.
-static guint64 browsers = 0U;
 static GMutex init_lock;
 static GCond init_cond;
 
@@ -472,13 +465,6 @@ class BrowserClient :
       src->state = CEF_SRC_CLOSED;
       g_cond_signal (&src->state_cond);
       g_mutex_unlock(&src->state_lock);
-      g_mutex_lock(&init_lock);
-      g_assert (browsers > 0);
-      browsers -= 1;
-      if (browsers == 0) {
-        CefQuitMessageLoop();
-      }
-      g_mutex_unlock (&init_lock);
     }
 
     // CefRequestHandler methods:
@@ -520,10 +506,6 @@ class BrowserClient :
         nullptr,
         nullptr
       );
-      g_mutex_lock (&init_lock);
-      g_assert (browsers < G_MAXUINT64);
-      browsers += 1;
-      g_mutex_unlock(&init_lock);
 
       browser->GetHost()->SetAudioMuted(true);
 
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
cc @aiden-jeffrey :)

aiden-jeffrey:
Moving comment from PR:

Multiple pipelines just aren't really supported the way this element is currently written. If you run two cefsrc elements at the same time (without calling CefShutdown), Initialize never gets acalled, which means that the second cefsrc just has the properties of the first.

aiden-jeffrey:
To be clearer - OnBeforeCommandLineProcessing doesn't get called - which is what maps many (or all, not checked) of the gst properties into cef land.

MathieuDuponchelle:
Oh, `OnBeforeCommandLineProcessing` is only called once too ok, annoying. I suppose all affected properties should be turned into environment variables instead, to make it clearer that they are per process.

aiden-jeffrey (approved):
This LGTM from my perspective and specific issue.

MathieuDuponchelle:
> I'm happy to take that PR if you need.

That'd be nice yes, by the way the idea would be to simply deprecate the properties, so if they're set and the environment variable is not the property value should still be used :)

amyspark:
@MathieuDuponchelle Some lingering questions now that I've caught up on this:

- are there any global properties that can be changed between CEF "initializations"?
- I think that there was an event in Gst land that was triggered on loop exit, on which we hook the CefShutdown task. How are pipelines being created after the loop's exited?


>  I think the issue here is CefShutdown is getting called, which according to the header should only be called on application exit. I've removed that, but now can't get the MakeBrowser CefTask to run in the second pipeline.

If the app's marked as "to be shut down", then the element will not respond to the second pipeline, which leads me to item (2) above.

aiden-jeffrey:
Just noticing now, but when you have multiple pipelines, it looks like the first frame on subsequent runs has some data left over in the buffer from the previous instance.

aiden-jeffrey:
> I'm happy to take that PR if you need.

https://github.com/centricular/gstcefsrc/pull/91

amyspark:
@aiden-jeffrey Hi, one question. In which operating system are you experiencing the crash? I've looked in the comments across the PRs but can't find the specifics on how to reproduce the issue you've been facing.

aiden-jeffrey:
> @aiden-jeffrey Hi, one question. In which operating system are you experiencing the crash? I've looked in the comments across the PRs but can't find the specifics on how to reproduce the issue you've been facing.

Ubuntu 24.04.

MathieuDuponchelle:
@amyspark yes unfortunately, many of the properties only matter at `CefInitialize` time, which is why we'll be moving to environment variables in #91 . Even if initialize / shutdown / initialize again worked, the problem still existed for concurrent instances anyway.

amyspark:
@MathieuDuponchelle All good. Maybe we should report these upstream (Google), it's clear that the engine as-is does not have a clean lifecycle.

aiden-jeffrey:
> @MathieuDuponchelle All good. Maybe we should report these upstream (Google), it's clear that the engine as-is does not have a clean lifecycle.

It's part of CEF's architecture that just one `CefInitialize` and one `CefShutdown` is [allowed](https://github.com/chromiumembedded/cef/blob/a231b7ead7e624dc5c2d79bf4c160aed9c8efeb9/include/cef_app.h#L98).

https://magpcss.org/ceforum/viewtopic.php?t=14648
https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19421
https://magpcss.org/ceforum/viewtopic.php?f=6&t=19158

I think the only way to support multiple CEF instances would be by re-writing `cefbin` to use something like https://gstreamer.freedesktop.org/documentation/ipcpipeline/index.html?gi-language=c.

MathieuDuponchelle:
> > @MathieuDuponchelle All good. Maybe we should report these upstream (Google), it's clear that the engine as-is does not have a clean lifecycle.
> 
> It's part of CEF's architecture that just one `CefInitialize` and one `CefShutdown` is [allowed](https://github.com/chromiumembedded/cef/blob/a231b7ead7e624dc5c2d79bf4c160aed9c8efeb9/include/cef_app.h#L98).
> 
> https://magpcss.org/ceforum/viewtopic.php?t=14648 https://www.magpcss.org/ceforum/viewtopic.php?f=6&t=19421 https://magpcss.org/ceforum/viewtopic.php?f=6&t=19158
> 
> I think the only way to support multiple CEF instances would be by re-writing `cefbin` to use something like https://gstreamer.freedesktop.org/documentation/ipcpipeline/index.html?gi-language=c.

Realistically, this would be up to the application that needs multiple instances to spawn one process per cefsrc and use some ipc mechanism to bring their output into a common pipeline if needed, I don't think we want to  take on that complexity in `gstcef` :)

PR #91, opened 2024-10-04
properties: Move gpu and chrome_extra_flags to env vars

aiden-jeffrey:
See https://github.com/centricular/gstcefsrc/pull/90#issuecomment-2393937044

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,26 @@ GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 \
 gst-launch-1.0 playbin uri=web+https://www.soundcloud.com/platform/sama
 ```
 
+### Note on Global CEF Parameters
+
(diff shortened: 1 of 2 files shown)


MathieuDuponchelle:
Nice, I'll try and make time to look in detail during the hackfest, do we need the same thing for `sandbox` and `js_flags`? I see that those are used for CEF initialization as well :/

aiden-jeffrey:
> Nice, I'll try and make time to look in detail during the hackfest, do we need the same thing for `sandbox` and `js_flags`? I see that those are used for CEF initialization as well :/

Ah good point, I'll go though things carefully tomorrow.

MathieuDuponchelle:
Looks good to me, that's annoyingly close to all the properties besides `url` :( Can you just add documentation for all the environment variables to a dedicated section in the README? Thanks again!

MathieuDuponchelle:
Perfect, thanks again :)

PR #92, opened 2024-10-31
Fix macOS and Linux build

amyspark:
Hi all,

This PR fixes two issues I was made aware of today:

- the macOS sandbox logic was being mistakenly built on Linux
- the event loop handler was renamed in 1931ae13940c5cb82ee3a07d5fed914ecf9b92c6, but that was not carried over to the state change handler for macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,9 +84,13 @@ set(GSTCEF_SRCS
 )
 
 set(GSTCEFSUBPROCESS_SRCS
-  gstcefloader.cc
   gstcefsubprocess.cc
 )
+
+if (APPLE)
+list(APPEND GSTCEF_SRCS gstcefloader.cc)
+list(APPEND GSTCEFSUBPROCESS_SRCS gstcefloader.cc)
+endif()
 
 pkg_check_modules(GST REQUIRED gstreamer-1.0
                   gstreamer-video-1.0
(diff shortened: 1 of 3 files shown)


MathieuDuponchelle:
> the event loop handler was renamed in https://github.com/centricular/gstcefsrc/commit/1931ae13940c5cb82ee3a07d5fed914ecf9b92c6, but that was not carried over to the state change handler for macOS

oops :|

MathieuDuponchelle:
lgtm, thanks :)

CMakeLists.txt
```diff
@@ -84,10 +84,14 @@ set(GSTCEF_SRCS
 )
```
vivia, line 84: Just one line above there's another `gstcefloader.cc`

gstcefsubprocess.cc
```diff
@@ -141,7 +140,7 @@ int main(int argc, char * argv[])
   CefMainArgs args(argc, argv);
 #endif
 
-#ifdef GST_CEF_USE_SANDBOX
+#if defined(__APPLE__) && defined(GST_CEF_USE_SANDBOX)
```
vivia, line 143: The same change has to be made also in `gstcef.cc` in two places

amyspark:
Filed https://github.com/centricular/gstcefsrc/pull/93 for @vivia's comments

PR #94, opened 2024-10-31
CMakeLists: Don't put target .so files in subdirectory

vivia:
Otherwise the plugin will fail to load


MathieuDuponchelle:
The plugin fails to load when installed right? I don't think I've ever tested this to be honest :)

MathieuDuponchelle:
@vivia I'll take the emoji reaction as confirmation then :P

PR #95, opened 2024-11-20
cefsrc: Clean up CEF object refs

aiden-jeffrey:
First raised here - https://github.com/centricular/gstcefsrc/pull/90#issuecomment-2393973894 - I am seeing some buffer corruption in the first `OnPaint` call when running multiple cefsrc elements.

It's not super deterministic, but looks like this:
image

with the next frame being a successful paint:
image

I think the issue is possibly a CEF one (see https://magpcss.org/ceforum/viewtopic.php?f=6&t=18906), but this branch is me trying to improve the Cef object cleanup (to no avail - doesn't fix the issue, but pushing to get your thoughts).

I wonder if we should just drop the first frame / OnPaint call?

Commit msg:
We were never setting src->app, so I removed it from the struct... Also I added the BrowserClient to _GstCefSrc, so that I could try and ensure it is getting dereferenced.

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -15,7 +15,6 @@
 #include <include/base/cef_bind.h>
 #include <include/base/cef_callback_helpers.h>
 #include <include/wrapper/cef_closure_task.h>
-#include <include/wrapper/cef_message_router.h>
 
 #include "gstcefsrc.h"
 #include "gstcefaudiometa.h"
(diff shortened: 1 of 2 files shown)


gstcefsrc.cc
```diff
@@ -967,6 +967,8 @@ gst_cef_src_stop (GstBaseSrc *base_src)
 {
   GstCefSrc *src = GST_CEF_SRC (base_src);
 
+  GST_OBJECT_LOCK (src);
```
MathieuDuponchelle, line 970: Why is this needed? Taking the object lock then waiting is a bit awkward, it could mean  an for instance an application potentially blocking while trying to set a property from its main thread

MathieuDuponchelle:
> I wonder if we should just drop the first frame / OnPaint call?

I think it would be best to first check with upstream before adding workarounds in the element.

aiden-jeffrey:
Fixed in https://github.com/centricular/gstcefsrc/pull/99

PR #99, opened 2025-08-28
Fix first frame corruption

aiden-jeffrey:
Reported [here](https://github.com/centricular/gstcefsrc/pull/95), this change makes sure to clear the buffer on the state change cb (which is what is currently being used to trigger all initialisation of the element).


MathieuDuponchelle:
can you rebase this? :)

Issue #84, opened 2024-05-21
Ideas on making src element only generate frames when a fresh buffer is created


aiden-jeffrey:
I'm looking to add a `vsync` type functionality to the cefsrc element that would only push buffers when a fresh one is painted in the `RenderHandler.OnPaint` method. Ultimately I want to be able to record webgl applications that may have a variable frame rate into a constant frame rate video. In other words, I want one frame in my mp4 file per `requestAnimationFrame` in js land.

Currently it's clear that (baring some initial paints), there is 1 `OnPaint` call per animation frame.

I can sort of get there by controlling the duration and pts time in `gst_cef_src_create`, but I was wondering if you have some better idea. Is the answer something to do with making the element non-live for this use case?

MathieuDuponchelle:
Would a meta to tag "original" video frames as opposed to copied / made-up be enough for you? You could then discard things as you see fit. Another option of course would be a property that would cause the source to output buffers marked as gap buffers, then in the demuxer they would be transformed as straight gap events, with the audio buffers still being demuxed as normal.

aiden-jeffrey:
Mmm, yes I was thinking that my approach might mess up audio. Out of interest, from an architecture perspective, why was the `cefmux` element required? I.e. why doesn't the `cefsrc` just expose an audio pad as well? Is it standard to stream audio packets on a video/raw pad and then demux?

MathieuDuponchelle:
> I.e. why doesn't the cefsrc just expose an audio pad as well? Is it standard to stream audio packets on a video/raw pad and then demux?

No, it is not standard but a workaround for the fact that `GstBaseSrc` is designed to epose a single `Always` source pad.

The alternative solution is a wrapper bin, with one source per output stream and a shared context (in this case the CEF browser), but at the time this was implemented cef had no support for audio capture, and it was then easier to retrofit a demuxer to the initial implementation :)

PR #107, opened 2026-02-03
Use CEF-provided timestamps

vivia:
See commit messages

Creating PR for easier reviewing


gstcefdemux.cc
```diff
@@ -290,9 +280,43 @@ gst_cef_demux_sink_query (GstPad *pad, GstObject *parent, GstQuery *query)
...
+      } else {
+        caps = gst_caps_new_empty ();
+        n = gst_caps_get_size (downstream_caps);
+        for (i = 0; i < n; i++) {
+          GstStructure *s = gst_caps_get_structure (downstream_caps, i);
+          s = gst_structure_copy (s);
+          gst_structure_set_name (s, "application/x-cef");
+          gst_caps_append_structure (caps, s);
+        }
+      }
```
heftig, line 305: Shouldn't you check that you only copy structures that have the `video/x-raw` name before changing it?

gstcefdemux.cc
```diff
@@ -301,6 +325,56 @@ gst_cef_demux_sink_query (GstPad *pad, GstObject *parent, GstQuery *query)
...
+      downstream_caps = gst_pad_peer_query_caps(sink, NULL);
+      gst_object_unref (sink);
+
+      n = gst_caps_get_size (downstream_caps);
+      for (i = 0; i < n; i++) {
+        GstStructure *s = gst_caps_get_structure (downstream_caps, i);
+        s = gst_structure_copy (s);
+        gst_structure_set_name (s, "video/x-raw");
+        gst_caps_append_structure (caps, s);
+      }
```
heftig, line 356: Ditto here. I think you should skip structures that aren't named `application/x-cef` to make sure misusing the element fails negotiation and doesn't do something weird.

heftig (approved):
LGTM, thanks!

sdroege:
I didn't review this in detail (but helped writing part of that code), but I think this should get out of draft status because it fixes stuff (the timestamping was completely wrong before) and I don't think we really need to wait for the other PR first (which also has quite some open questions how to do things at all). @aiden-jeffrey ?

@MathieuDuponchelle What do you think?

gstcefsrc.cc
```diff
@@ -1,3 +1,4 @@
+#include "gst/gstinfo.h"
```
sdroege: This should be removed (need to configure clangd with `-header-insertion=never`)

MathieuDuponchelle:
@sdroege on my list of things to review, need to take some time to do that properly

gstcefdemux.cc
```diff
@@ -3,6 +3,7 @@
 #include "gstcefdemux.h"
 #include "gstcefaudiometa.h"
 
+#define CEF_SINK_CAPS "application/x-cef"
```
aiden-jeffrey: You probably need to update the readme as i don't think the current gst-launch commands listed there work any more. I had to replace `video/x-raw, width=1920, height=1080, framerate=60/1` with `application/x-cef`.

gstcefsrc.h
```diff
@@ -71,7 +71,7 @@ struct _GstCefSrc {
 
   GMutex queue_lock;
   GCond queue_cond;
-  GstVecDeque *queue;
+  GstQueueArray *queue;
```
MathieuDuponchelle: Can you perhaps add the deprecated TODO here for better visibility? Also perhaps fix up with the queue -> VecDeque commit :)

html/raf_test.html
```diff
@@ -0,0 +1,99 @@
+<html>
```
MathieuDuponchelle, line 1: We might want to mention those test pages in the README, perhaps by updating the example command lines?

gstcefsrc.cc
```diff
@@ -257,10 +258,20 @@ class RenderHandler : public CefRenderHandler
...
-      gst_buffer_replace (&(src->current_buffer), new_buffer);
-      gst_buffer_unref (new_buffer);
-      GST_OBJECT_UNLOCK (src);
+      GstClock *clock = gst_element_get_clock (GST_ELEMENT (src));
+      GstClockTime base_time = gst_element_get_base_time (GST_ELEMENT (src));
+      GstClockTime now_gst = gst_clock_get_time (clock);
+      gst_object_unref (clock);
+
+      // running time
+      GstClockTime gst_pts = (now_gst > base_time) ? now_gst - base_time : 0;
```
MathieuDuponchelle: why not use `gst_element_get_current_running_time()` ?

gstcefdemux.cc
```diff
@@ -3,7 +3,7 @@
 #include "gstcefdemux.h"
 #include "gstcefaudiometa.h"
 
-#define CEF_SINK_CAPS "application/x-cef"
+#define CEF_SINK_CAPS "application/x-cef, format=BGRA, width=[1, 2147483647], height=[1, 2147483647], framerate=[1/1, 60/1], pixel-aspect-ratio=1/1"
```
aiden-jeffrey: we should probably rename this to CEF_DEMUX_SINK_CAPS, for clarity.

gstcefdemux.cc
```diff
@@ -280,9 +280,43 @@ gst_cef_demux_sink_query (GstPad *pad, GstObject *parent, GstQuery *query)
...
-    case GST_QUERY_CAPS:
-      ret = gst_pad_peer_query(demux->vsrcpad, query);
+    case GST_QUERY_CAPS: {
+      gint i, n;
+      GstCaps *downstream_caps, *filter, *tmp;
+      GstCaps *caps;
+
+      downstream_caps = gst_pad_peer_query_caps(demux->vsrcpad, NULL);
+
+      if (gst_caps_is_any (downstream_caps)) {
```
aiden-jeffrey, line 294: is this right, should it not be `if (!gap_caps_is_any(...))`?

aiden-jeffrey (approved):
I agree with @sdroege, this looks ready to take out of WIP to me. It solves the same issue as https://github.com/centricular/gstcefsrc/pull/104 does but in a much better way. Great commit messages also.

gstcefsrc.cc
```diff
@@ -326,7 +337,39 @@ class AudioHandler : public CefAudioHandler
...
 
-    GST_OBJECT_LOCK (src);
+    GstClock *clock = gst_element_get_clock (GST_ELEMENT (src));
+    GstClockTime base_time = gst_element_get_base_time (GST_ELEMENT (src));
+    GstClockTime now_unix = g_get_real_time () * 1000; // nsec
+    GstClockTime now_gst = gst_clock_get_time (clock);
+    GstClockTime cef_pts_unix = pts * 1000000; // pts is in msec
+    GstClockTime capture_gst;
+
+    if (cef_pts_unix > now_unix) {
```
MathieuDuponchelle, line 345: As I understand this code the intent is to "ignore" potential drift, but I don't know if that's the optimal way to go about it, in particular which piece of the pipeline will eventually have to insert or drop samples to compensate (it might very well be the optimal way but I'm not familiar enough with audio aspects I'm afraid).

In my understanding the best way to go about this is to have the element provide a clock and calibrate it, @sdroege what's your opinion here?

gstcefsrc.cc
```diff
@@ -257,10 +258,20 @@ class RenderHandler : public CefRenderHandler
...
-      GST_OBJECT_UNLOCK (src);
+      GstClock *clock = gst_element_get_clock (GST_ELEMENT (src));
+      GstClockTime base_time = gst_element_get_base_time (GST_ELEMENT (src));
+      GstClockTime now_gst = gst_clock_get_time (clock);
+      gst_object_unref (clock);
+
+      // running time
+      GstClockTime gst_pts = (now_gst > base_time) ? now_gst - base_time : 0;
+
+      GST_BUFFER_PTS (new_buffer) = gst_pts;
```
MathieuDuponchelle, line 266: The old code added a duration according to the framerate, I take it you don't set one anymore because the element is now variable framerate? Should the caps reflect this? Doesn't that create a problem with the audio_events design, where audio might pile up if OnPaint isn't called?

gstcefsrc.cc
```diff
@@ -885,6 +934,19 @@ gst_cef_src_change_state(GstElement *src, GstStateChange transition)
   if (result == GST_STATE_CHANGE_FAILURE) return result;
   result = GST_ELEMENT_CLASS(parent_class)->change_state(src, transition);
 
+  switch (transition) {
+    case GST_STATE_CHANGE_PAUSED_TO_READY: {
+      GstCefSrc *self = GST_CEF_SRC (src);
+
+      g_mutex_lock (&self->queue_lock);
+      g_queue_clear_full (self->queue, (GDestroyNotify) gst_buffer_unref);
```
MathieuDuponchelle: Is it not enough to do this in `stop` ?

gstcefdemux.cc
MathieuDuponchelle: What's the intent here? :) A nice effect of using video/x-raw caps was that you could do away with the demuxer entirely if you wanted :)

gstcefsrc.cc
```diff
@@ -707,7 +708,9 @@ static GstFlowReturn gst_cef_src_create(GstPushSrc *push_src, GstBuffer **buf)
     src->audio_events = NULL;
   }
 
-  while (g_queue_is_empty(src->queue) && !src->flushing) {
+  while (g_queue_is_empty(src->queue) &&
+         !src->flushing &&
+         (!src->audio_buffers || !src->downstream_demuxer)) {
```
MathieuDuponchelle, line 738: Ah I see OK, that's what I get for reviewing per-commit :) I wonder if we could still keep the video/x-raw caps if we set the appropriate flags on dummy buffers (and not worry that there's a demuxer downstream or not), eg `GST_BUFFER_FLAG_GAP`? Side note, I think we can do away with the `n_frames` field :)

html/raf_test.html
vivia: This one is for @aiden-jeffrey actually, I just rebased my commit on top of these pages!

gstcefsrc.cc
vivia: Oh you're right, `n_frames` was never read, thanks! Removed it :)

gstcefsrc.h
vivia: I'd rather leave the commit there on purpose so we can just revert it when it's time :)

html/raf_test.html
aiden-jeffrey: i can handle that in a fast follow if you like

vivia: That would be helpful, thanks! I'm soon-ish-ly going to push some fixes to the review comments and you can add that commit on top of mine :)

gstcefsrc.h
MathieuDuponchelle: makes sense :)

gstcefsrc.cc
sdroege: This is the (or a) correct way to timestamp media (audio or not) if the source is producing it based on a different clock than the pipeline clock. Downstream will have to deal with that, just like in many other situations. And audiorate/audiobuffersplit or actual audio sink or many other elements can take care of that.

The element could also provide a clock but that seems a bit pointless. The correct clock would be the real-time system clock, and if that's configured on the pipeline then then the calculations here would end up with approximately 0 drift already anyway.

gstcefsrc.cc
sdroege: I don't think this makes sense. A `GST_BUFFER_FLAG_GAP` buffer still needs to be valid (everything needs to behave the same even if downstream ignores the flag). You'd have to invent some kind of valid video buffer here then, which would e.g. be a black one or a copy of the previous video frame. That doesn't seem useful to do here as it changes behaviour for downstream if it doesn't know about the audio for no reason at all, and if downstream knows about the audio it would be properly demuxed inside the demuxer now already without introducing unnecessary video frames.

html/raf_test.html
vivia: @aiden-jeffrey Changes pushed, it's all yours now! :) Thanks!

gstcefsrc.cc
MathieuDuponchelle: @vivia my question about the framerate still needs resolution, I think you need to shift to 0/1 :)

gstcefsrc.cc
MathieuDuponchelle: OK, thanks for clarifying this :) CEF being software indeed points to it using the system time, good point re custom clock being redundant :)

gstcefsrc.cc
MathieuDuponchelle: OK, the nice side effect here is of course that this should (partly) resolve https://github.com/centricular/gstcefsrc/pull/105 / https://github.com/centricular/gstcefsrc/issues/84 , I think we'll just need to expose an is-live property to get the expected behavior :) We do need to advertise a variable framerate though, not reopening this because I already unresolved another thread for that discussion :)

gstcefsrc.cc
sdroege: It's even documented that it uses the real-time system clock.

gstcefsrc.cc
sdroege: A non-live mode for cefsrc needs a lot more thought

gstcefsrc.cc
MathieuDuponchelle: where ooc?

gstcefsrc.cc
MathieuDuponchelle: That will also mean we need to expose a different method for configuring CEF's framerate (see `SetWindowlessFrameRate` call)

gstcefsrc.cc
MathieuDuponchelle: What issue do you foresee? Now that I think of it if the browser is rendering slower than real time our timestamp calculation will obviously be thrown off, I wonder what timestamps CEF will give us in that case :)

gstcefsrc.cc
sdroege: Don't ask me things from 4 months ago :D Somewhere in the docs are headers!

gstcefsrc.cc
sdroege: For video you don't get timestamps and it's unclear what it will even render if it's not operating in real-time. Probably depends completely on the content, whether that is doing anything time based or frame based?

For audio, you get data according to the system clock. I don't think this can possibly work non-live at all.

The whole rendering model of CEF is based on real-time output AFAICS.

MathieuDuponchelle: OK, I could swear @aiden-jeffrey's use case was the ability to use CEF as non-realtime renderer of sorts, but maybe that's indeed not possible :)

sdroege: Yes, we discussed that 4 months ago on Matrix and came to the conclusion that for audio we have no answer and for video maybe there's something special possible but unclear.

aiden-jeffrey: Yes it is possible, but you have to combine it with some magic on the browser side, making requestAnimationFrame think it's running at a constant frame rate. Audio is another question, I think (at least for now) we'd just accept that we can't generate constant frame rate videos with audio.

MathieuDuponchelle: right, separate concern anyway

aiden-jeffrey: I will update 105 once this PR is merged and we can discuss there I suppose.

aiden-jeffrey: I will add all the "magic" to this repo as something people can use on the web side.

sdroege: Yes that makes sense to me

html/raf_test.html
aiden-jeffrey: I'll update the docs tomorrow first thing.

gstcefsrc.cc
sdroege: I think providing the framerate in the caps like before still makes sense. It's only an indication and in perfect conditions CEF will produces frames with that framerate. That's more than what you can say about `v4l2src`

MathieuDuponchelle: >  It's only an indication and in perfect conditions CEF will produces frames with that framerate.

Is that the case though? As far as I understand CEF may not produce frames at all for an extended time with static pages now

sdroege: Ah I thought with `SetWindowlessFrameRate()` it would produce frames in that framerate if it can

MathieuDuponchelle: The doc says:

>  Set the maximum rate in frames per second (fps) that CefRenderHandler:: OnPaint will be called for a windowless browser. The actual fps may be lower if the browser cannot generate frames at the requested rate. The minimum value is 1 and the maximum value is 60 (default 30). Can also be set at browser creation via CefBrowserSettings.windowless_frame_rate. 

So actually unclear OnPaint will still get called when nothing has changed. @vivia can you check this with eg a blank page? If indeed the only reason why OnPaint might not get called at the specified framerate is that the browser can't keep up, then I agree we should keep things as is, otherwise I think variable framerate is the correct thing to advertise :)

vivia: I tested with some static pages. OnPaint isn't called exactly once at the beginning, but also nowhere near the specified framerate. So yes, I think variable framerate is the correct thing to advertise, but we'd still need a way to specify the maximum framerate, most likely a property. Currently this would be a very breaking change though, because existing pipelines that specify a framerate would fail to negotiate, they'd need to specify it with the property instead, and if they want the full framerate they'd need a `livesync` downstream.

I mean I can still do it that way, but I'd need you to sign off on this breaking change :)

gstcefsrc.cc
vivia: I moved this part of the code before the parent class change_state because of the deadlock. I mean it works this way, so I don't want to risk it deadlocking again, what do you think? 😅

gstcefdemux.cc
vivia: The intent is to know whether we have a demuxer downstream, and if we do we use application/x-cef, otherwise we use video/x-raw. Leaving this unresolved because I will still need to update the readme, but for that I'd need to know whether I'll break the API :)

gstcefsrc.cc
MathieuDuponchelle: Thanks for testing!

A property was also my suggestion yes, and maybe let's add a note to the README (and check that the example pipelines still work).

As for the change being breaking, so be it, I actually like it better as it makes the behavior change explicit :)

gstcefsrc.cc
MathieuDuponchelle: Ah I see, I think the correct fix is to implement `https://gstreamer.freedesktop.org/documentation/base/gstbasesrc.html?gi-language=c#GstBaseSrcClass::unlock` :)

MathieuDuponchelle: (and `unlock_stop()`)

aiden-jeffrey (changes requested):
There's some major issue with this change as far as I can tell. Compare this pipeline on master vs this branch:

`GST_CEF_CHROME_EXTRA_FLAGS="autoplay-policy=no-user-gesture-required"  GST_CEF_GPU_ENABLED=true  gst-launch-1.0 cefbin name=cef cefsrc::url=https://webglsamples.org/aquarium/aquarium.html  cefsrc::chrome-extra-flags="autoplay-policy=no-user-gesture-required" cef.video ! video/x-raw, width=1920, height=1080, framerate=25/1 ! videoconvert ! autovideosink cef.audio ! audioconvert ! audio/x-raw, format=S32LE, channels=2 ! audioresample ! audio/x-raw, rate=48000 ! fakesink`

You should see significant memory leakage and terrible playback performance in this timestamp branch.

We also see this same behaviour when using the `cefsrc` and `cefdemux` elements directly.

Memory usage graph for the above pipeline:
mem

aiden-jeffrey:
And for reference, the same pipeline on master:

mem-master

aiden-jeffrey:
> There's some major issue with this change as far as I can tell. Compare this pipeline on master vs this branch:

I don't get the same memory leakage (albeit the pipeline does struggle to go to playing) rendering example.com.

aiden-jeffrey:
I actually see this issue on the test page too (`cd html && python -m http.server 9000`):

`GST_CEF_CHROME_EXTRA_FLAGS="autoplay-policy=no-user-gesture-required"  gst-launch-1.0 cefbin name=cef cefsrc::url=http://localhost:9000/raf_test.html?something=food  cefsrc::chrome-extra-flags="autoplay-policy=no-user-gesture-required" cef.video ! video/x-raw, width=1920, height=1080, framerate=25/1 ! videoconvert ! autovideosink cef.audio ! audioconvert ! audio/x-raw, format=S32LE, channels=2 ! audioresample ! audio/x-raw, rate=48000 ! fakesink`

Memory plot:
mem

MathieuDuponchelle:
@aiden-jeffrey given the size of the leaks I expect this is just an extra buffer reference if you feel like debugging this :)

gstcefdemux.cc
```diff
@@ -263,10 +243,20 @@ gst_cef_demux_sink_event (GstPad *pad, GstObject *parent, GstEvent *event)
...
+      GstCaps *caps;
+
+      gst_event_parse_caps (event, &upstream_caps);
+      s = gst_structure_copy (gst_caps_get_structure (upstream_caps, 0));
+      gst_structure_set_name (s, "video/x-raw");
+      caps = gst_caps_new_full (s, NULL);
+      demux->vcaps_event = gst_event_new_caps (caps);
+      gst_caps_unref (caps);
       demux->need_caps = TRUE;
       event = NULL;
```
sdroege, line 262: This is leaking the caps event

gstcefdemux.cc
```diff
@@ -290,9 +280,43 @@ gst_cef_demux_sink_query (GstPad *pad, GstObject *parent, GstQuery *query)
...
 
   switch (GST_QUERY_TYPE (query)) {
-    case GST_QUERY_CAPS:
-      ret = gst_pad_peer_query(demux->vsrcpad, query);
+    case GST_QUERY_CAPS: {
+      gint i, n;
+      GstCaps *downstream_caps, *filter, *tmp;
+      GstCaps *caps;
+
+      downstream_caps = gst_pad_peer_query_caps(demux->vsrcpad, NULL);
```
sdroege, line 292: These caps are leaked

gstcefdemux.cc
```diff
@@ -301,6 +325,56 @@ gst_cef_demux_sink_query (GstPad *pad, GstObject *parent, GstQuery *query)
...
+
+  switch (GST_QUERY_TYPE (query)) {
+    case GST_QUERY_CAPS: {
+      gint i, n;
+      GstCaps *downstream_caps, *filter, *tmp;
+      GstCaps *caps = gst_caps_new_empty();
+      GstPad *sink = gst_element_get_static_pad (GST_ELEMENT (parent), "sink");
+
+      g_assert_nonnull (sink);
+      downstream_caps = gst_pad_peer_query_caps(sink, NULL);
```
sdroege, line 347: Leaked

gstcefsrc.cc
```diff
@@ -1049,11 +1109,22 @@ gst_cef_src_fixate (GstBaseSrc * base_src, GstCaps * caps)
...
     gst_structure_set (structure, "framerate", GST_TYPE_FRACTION, DEFAULT_FPS_N, DEFAULT_FPS_D, nullptr);
 
+  intersected_caps = gst_caps_intersect (caps, application_cef_caps);
+
+  if (!gst_caps_is_empty (intersected_caps)) {
+    src->downstream_demuxer = TRUE;
+    gst_caps_unref (caps);
+    caps = intersected_caps;
+  } else {
+    src->downstream_demuxer = FALSE;
```
sdroege, line 1136: Leaking the empty caps here

sdroege (commented):
Found no buffer leak but some small caps/event leaks. I hope these are not the reason

gstcefdemux.cc
```diff
@@ -263,10 +243,20 @@ gst_cef_demux_sink_event (GstPad *pad, GstObject *parent, GstEvent *event)
...
+    case GST_EVENT_CAPS: {
+      GstCaps *upstream_caps;
+      GstStructure *s;
+      GstCaps *caps;
+
+      gst_event_parse_caps (event, &upstream_caps);
+      s = gst_structure_copy (gst_caps_get_structure (upstream_caps, 0));
+      gst_structure_set_name (s, "video/x-raw");
+      caps = gst_caps_new_full (s, NULL);
+      demux->vcaps_event = gst_event_new_caps (caps);
```
sdroege, line 258: Should probably make sure to not leak any event that might've been here before

gstcefsrc.cc
```diff
@@ -254,13 +254,18 @@ class RenderHandler : public CefRenderHandler
...
-      gst_buffer_replace (&(src->current_buffer), new_buffer);
-      gst_buffer_unref (new_buffer);
-      GST_OBJECT_UNLOCK (src);
+      // running time
+      GstClockTime gst_pts = gst_element_get_current_running_time (GST_ELEMENT (src));
+
+      GST_BUFFER_PTS (new_buffer) = gst_pts;
+
+      g_mutex_lock (&src->queue_lock);
+      gst_queue_array_push_tail (src->queue, new_buffer);
```
sdroege, line 269: Maybe we're producing buffers faster than they're consumed? Should maybe add a limit here and start dropping buffers at some point (oldest first) once the queue becomes too big

gstcefsrc.cc
```diff
@@ -335,7 +372,8 @@ class AudioHandler : public CefAudioHandler
     }
 
     gst_buffer_list_add (src->audio_buffers, buf);
```
sdroege, line 378: Same here, this potentially accumulates infinite audio buffers

aiden-jeffrey:
Thanks @sdroege & @MathieuDuponchelle - i'll have a go at resolving this tomorrow.

gstcefsrc.cc
aiden-jeffrey: Both this and the audio buffer queue are accumulating buffers infinitely. See [my branch](https://github.com/vivia/gstcefsrc/pull/1) against this PR with extra logging. In [my earlier attempt to solve this problem](https://github.com/centricular/gstcefsrc/pull/104),  I had to add a certain amount of delay to the audio pts - maybe something similar is required here. I'm doubtful that we want to drop buffers here, but I think instead we should understand what's happening.

gstcefdemux.cc
aiden-jeffrey: Resolved these here: https://github.com/vivia/gstcefsrc/pull/1

gstcefsrc.cc
MathieuDuponchelle: What causes the build up? In your example pipeline the sinks are both synchronized to the pipeline clock, does this mean cefsrc queues buffers faster than real time?

gstcefsrc.cc
vivia: Oh thanks, I didn't know that API! Fixed :)

gstcefsrc.cc
vivia: Done, thanks! :)

gstcefsrc.cc
aiden-jeffrey: Not sure, didn't have loads of time to investigate.

aiden-jeffrey:
@vivia - with your latest framerate changes, the cefbin is no longer functional. You need to update the `CEF_VIDEO_CAPS` in gstcefbin.cc, but even when you do that the README pipeline fails to negotiate:

```
$ GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0     cefbin name=cef cefsrc::url="https://soundcloud.com/platform/sama"     cef.video ! video/x-raw, width=1920, height=1080 ! videoconvert ! xvimagesink     cef.audio ! audioconvert ! audiomixer ! autoaudiosink

WARNING: erroneous pipeline: could not link cef to videoconvert0, cef can't handle caps video/x-raw, width=(int)1920, height=(int)1080
```

vivia:
> @vivia - with your latest framerate changes, the cefbin is no longer functional. You need to update the `CEF_VIDEO_CAPS` in gstcefbin.cc, but even when you do that the README pipeline fails to negotiate:
> 
> ```
> $ GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0     cefbin name=cef cefsrc::url="https://soundcloud.com/platform/sama"     cef.video ! video/x-raw, width=1920, height=1080 ! videoconvert ! xvimagesink     cef.audio ! audioconvert ! audiomixer ! autoaudiosink
> 
> WARNING: erroneous pipeline: could not link cef to videoconvert0, cef can't handle caps video/x-raw, width=(int)1920, height=(int)1080
> ```

Thanks for the feedback. I made the change and it works, weird.

```
vivia@iapetos ~/g/M/gstcefsrc (audiopts-cef)> git diff --no-ext-diff
diff --git a/gstcefbin.cc b/gstcefbin.cc
index 0cc6743..6867154 100644
--- a/gstcefbin.cc
+++ b/gstcefbin.cc
@@ -1,7 +1,7 @@
 #include "gstcefbin.h"
 #include "gstcefaudiometa.h"
 
-#define CEF_VIDEO_CAPS "video/x-raw, format=BGRA, width=[1, 2147483647], height=[1, 2147483647], framerate=[1/1, 60/1], pixel-aspect-ratio=1/1"
+#define CEF_VIDEO_CAPS "video/x-raw, format=BGRA, width=[1, 2147483647], height=[1, 2147483647], framerate=0/1, pixel-aspect-ratio=1/1"
 #define CEF_AUDIO_CAPS "audio/x-raw, format=F32LE, rate=[1, 2147483647], channels=[1, 2147483647], layout=interleaved"
 
 static GstURIType
vivia@iapetos ~/g/M/gstcefsrc (audiopts-cef)> gst-launch-1.0 cefbin name=cef cefsrc::url="https://soundcloud.com/platform/sama"     cef.video ! video/x-raw, width=1920, height=1080 ! videoconvert ! xvimagesink     cef.audio ! audioconvert ! audiomixer ! autoaudiosink
Setting pipeline to PAUSED ...
[0327/171833.917615:WARNING:cef/libcef/common/resource_util.cc:83] Please customize CefSettings.root_cache_path for your application. Use of the default value may lead to unintended process singleton behavior.
Pipeline is live and does not need PREROLL ...
Progress: (open) Creating CEF browser client
Progress: (open) Creating CEF browser ...
Progress: (open) Waiting for CEF browser initialization...
Progress: (open) CEF browser created
Pipeline is PREROLLED ...
Setting pipeline to PLAYING ...
New clock: GstSystemClock
Redistribute latency...
Redistribute latency...
Redistribute latency...
[3664726:3664726:0327/171834.368369:ERROR:gpu/command_buffer/service/gles2_cmd_decoder_passthrough.cc:1096] [GroupMarkerNotSet(crbug.com/242999)!:B02DA51949560000]Automatic fallback to software WebGL has been deprecated. Please use the --enable-unsafe-swiftshader (about:flags#enable-unsafe-swiftshader) flag to opt in to lower security guarantees for trusted content.
[3664726:3664726:0327/171834.868994:ERROR:gpu/command_buffer/service/gles2_cmd_decoder_passthrough.cc:1096] [GroupMarkerNotSet(crbug.com/242999)!:702DA51949560000]Automatic fallback to software WebGL has been deprecated. Please use the --enable-unsafe-swiftshader (about:flags#enable-unsafe-swiftshader) flag to opt in to lower security guarantees for trusted content.
^Chandling interrupt.
Interrupt: Stopping pipeline ...
Execution ended after 0:00:01.619932146
Setting pipeline to NULL ...
Freeing pipeline ...
```

aiden-jeffrey:
> Thanks for the feedback. I made the change and it works, weird.

You're right, sorry. This branch still leaks memory with this pipeline:
```
GST_CEF_CHROME_EXTRA_FLAGS="autoplay-policy=no-user-gesture-required"  GST_CEF_GPU_ENABLED=true  GST_PLUGIN_PATH=Release:$GST_PLUGIN_PATH gst-launch-1.0 cefbin name=cef cefsrc::url=https://webglsamples.org/aquarium/aquarium.html  cefsrc::chrome-extra-flags="autoplay-policy=no-user-gesture-required" cef.video ! video/x-raw, width=1920, height=1080 ! videoconvert ! autovideosink cef.audio ! audioconvert ! audio/x-raw, format=S32LE, channels=2 ! audioresample ! audio/x-raw, rate=48000 ! fakesink
```

Interestingly the test app variant of the above pipeline no longer leaks:
```
GST_CEF_CHROME_EXTRA_FLAGS="autoplay-policy=no-user-gesture-required"  gst-launch-1.0 cefbin name=cef cefsrc::url=http://localhost:9000/raf_test.html?something=food  cefsrc::chrome-extra-flags="autoplay-policy=no-user-gesture-required" cef.video ! video/x-raw, width=1920, height=1080, framerate=25/1 ! videoconvert ! autovideosink cef.audio ! audioconvert ! audio/x-raw, format=S32LE, channels=2 ! audioresample ! audio/x-raw, rate=48000 ! fakesink
```

vivia:
That's strange, I can't reproduce the leak here. Can you debug this a bit further using the logs you added? Is it indeed buffers piling up internally before they have the chance to be consumed?

ETA: Also my changes shouldn't have had any effect on the pipeline's PLAYING state, I just fixed a negotiation issue and cleaned up the shutdown path.

vivia:
@aiden-jeffrey Sorry for the long absence, I had other fires to put out in the meantime.

So I tried exactly your aquarium pipeline and it doesn't preroll. Looks like CEF won't produce any audio buffers when the webpage is silent. I added back the audiomixer in the pipeline (which is still described in the README, fortunately) and ran it for several minutes, in different permutations, without noticing any leak. I wonder how yours managed to preroll at the first place.

I'm not happy with the fact that the naïve pipeline doesn't preroll, but well, we still have the README describing that you should use an audiomixer, so I guess it can be kept like that.

MathieuDuponchelle:
@vivia sorry hadn't noticed your updates, what's the status here? Are you satisfied with the updated state?

vivia:
@MathieuDuponchelle Yes, it seems to be running fine for us!

I wonder if @aiden-jeffrey 's reported leak is another case of https://blog.nirbheek.in/2026/05/an-esoteric-type-of-memory-leak.html , in which case we'll have to think about when/where to call `malloc_trim` - can/should we really do it on every buffer? A counter for every N-th buffer?

MathieuDuponchelle:
> can/should we really do it on every buffer? A counter for every N-th buffer?

I wonder if there's a way to monitor fragmentation, but we would also need to confirm that this is indeed the problem :) @aiden-jeffrey can you perhaps do some A/B testing? I also don't really like the idea of a specific plugin triggering such a process-wide mechanism :|

MathieuDuponchelle:
I'll merge this now regardless, it's a very needed improvement, thanks again :)

sdroege:
That's IMHO something to be done at the application level based on the application's knowledge about allocation patterns. For example the application could do that whenever a cefsrc is destroyed, or whenever a pipeline is.

PR #110, opened 2026-08-11
macos: fix build against current CEF and clang

wagenet:
## Summary

The macOS build is currently broken against current CEF and current clang. This fixes it; no behaviour changes on Linux or Windows.

Everything here is guarded by `if (NOT APPLE)` / `#ifdef __APPLE__`, or is a format-string correction that is a no-op on LP64 Linux.

## The problems

**1. `libcef_lib` does not exist on macOS — configure fails outright.**

`ADD_LOGICAL_TARGET("libcef_lib", ...)` and the two `target_link_libraries(... libcef_lib ...)` calls reachable on Apple are unconditional. But CEF's `cef_variables.cmake` defines `CEF_LIB_DEBUG`/`CEF_LIB_RELEASE` only under `OS_LINUX` (`libcef.so`) and `OS_WINDOWS` (`libcef.lib`), and defines `CEF_SANDBOX_LIB_*` under neither. The macOS distribution ships only `Chromium Embedded Framework.framework` — there is no import library at all.

That is by design, and this project already accounts for it: `gstcefloader.cc` is compiled on Apple precisely so the framework can be `dlopen`ed at runtime via `cef_load_library()`, and macOS `CEF_STANDARD_LIBS` contains only system frameworks. So macOS should never link `libcef_lib`.

Both branches expand to empty paths, and `cmake_policy(SET CMP0111 NEW)` at the top of `CMakeLists.txt` turns that into a hard error:

```
CMake Error in CMakeLists.txt:
  IMPORTED_LOCATION or IMPORTED_IMPLIB not set for imported target
  "libcef_lib" configuration "Release".
```

(`-DUSE_SANDBOX=OFF` does not help — the `else` branch is equally undefined on macOS.)

**2. `CMAKE_OBJCXX_STANDARD` is 14, but CEF headers need C++17.**

`include/base/cef_bind.h` and friends use `std::void_t`, `std::conjunction`, `std::disjunction`, `std::bool_constant` and `std::in_place_t`. CEF itself compiles with `-std=c++17` (`cef_variables.cmake`), so this matches it.

**3. Undefined function-pointer cast.**

`init_cef` is `gpointer (*)(GstCefSrc *)`, cast to `dispatch_function_t` — `void (*)(void *)`. Calling through a mismatched function pointer type is UB, and clang now diagnoses it as `-Wcast-function-type-mismatch`, which `-Werror` makes fatal. Fixed with a correctly typed trampoline rather than by widening the cast or suppressing the warning.

**4. `GstClockTime` format strings.**

`GstClockTime` is `guint64`. On LP64 Linux that is `unsigned long`, so `%lu` happens to be correct; on macOS it is `unsigned long long` and `-Wformat` rejects it. `G_GUINT64_FORMAT` is correct everywhere.

## Testing

Built and run on macOS arm64 (Apple silicon), GStreamer 1.28.6, CEF `144.0.21+g4f5b28c+chromium-144.0.7559.248`, on top of `c89c2ce`:

- Configures and builds clean with upstream's `-Werror`, no suppressions.
- `cefsrc`, `cefdemux` and `cefbin` all register.
- Renders correctly: pages produce BGRA frames with correct content.
- **Per-pixel alpha is correct** — a page with `background: transparent` and a red `<div>` yields exactly `a=255` inside the div and `a=0` outside (`(0,0,0,0)`), which is what makes it usable as a broadcast overlay.

Also exercised in a real application (Eyevinn Strom) compositing a transparent CEF page as a downstream keyer over a program feed.

Note for anyone reproducing: on macOS CEF refuses to initialise unless the host process is inside an `.app` bundle — an unbundled `gst-launch-1.0` hangs silently at PAUSED. The README's bundle instructions are required, not optional.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,10 @@ project(gstcef)
 
 if (APPLE)
   enable_language(CXX OBJCXX)
-  set(CMAKE_OBJCXX_STANDARD 14)
+  # CEF 144 headers (base/cef_bind.h et al) use C++17 library features --
+  # std::void_t, conjunction, disjunction, bool_constant, in_place_t. CEF itself
+  # builds with -std=c++17 (cef_variables.cmake:290), so match it here.
+  set(CMAKE_OBJCXX_STANDARD 17)
   set(CMAKE_OBJCXX_STANDARD_REQUIRED ON)
 endif()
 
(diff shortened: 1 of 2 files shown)


MathieuDuponchelle:
@amyspark do you have some time to review this? :)

wagenet:
I should say that while this is with Claude, I reviewed the code and, while I'm not a C expert, it seems reasonable. I've also validated it locally on my Mac.

amyspark:
Sorry for the delay. LGTM

amyspark (approved):

MathieuDuponchelle:
thanks @amyspark :)

PR #111, opened 2026-08-11
cefsrc: Bound the frame and audio queues

wagenet:
`RenderHandler::OnPaint` allocates a full frame and appends it to `src->queue` on every paint. Since #112 it discards frames in PAUSED, but in PLAYING it still queues unconditionally, and the only consumer is `gst_cef_src_create()` on the streaming task. If that task stops while the element stays in PLAYING, the queue grows by `width * height * 4` bytes per paint (8.3 MB at 1080p) until the process runs out of memory.

We hit this when downstream caps pinned a framerate that `cefsrc`'s `0/1` could not satisfy: `gst_base_src_loop` paused the task on `GST_FLOW_NOT_NEGOTIATED`, the pipeline stayed in PLAYING, and the browser kept painting. A leaky queue downstream does not prevent this, since `create()` is not being called and the queue forwards `NOT_NEGOTIATED` upstream anyway. On macOS CEF installs PartitionAlloc process-wide, so the failed allocation aborts inside `OnPaint` instead of returning NULL.

## Changes

- Cap the frame queue at 2 and drop the oldest frame. In steady state `create()` pops one frame per paint, so the cap only engages once draining has stopped.
- Cap `src->audio_buffers` at 128 and drop the oldest packets. It is drained by the same `create()` call and normally holds a handful.

The commits are independent. #105 is orthogonal: with it the queue stays at one frame and this cap never engages.

## Testing

macOS/arm64, CEF 144.0.21, GStreamer 1.28.6; this branch and master built side by side. The page animates a 1920×1080 canvas and plays a WebAudio tone. Memory is `phys_footprint`, sampled once a second.

**Task stopped in PLAYING.** `cefsrc ! video/x-raw,framerate=30/1 ! fakesink` fails with `NOT_NEGOTIATED` while the element stays in PLAYING.

| build | duration | max frame queue | max audio list | footprint |
|---|---|---|---|---|
| master | 15 s (stopped early) | 446 | 638 | 58 → 3638 MB |
| this branch | 60 s | 2 | 128 | flat, 93–98 MB |

**Normal playback.** `cefsrc ! cefdemux` with both pads into `fakesink` for 30 s: 928 paints, frame queue and audio list never above 1, nothing dropped. `cefsrc ! fakesink` likewise dropped nothing.

Not run on Linux or Windows. The repo has no automated tests, so there is no regression test.

<details><summary>Crash evidence</summary>

Two aborts ~3h apart with identical stacks: `OnPaint → gst_buffer_new_allocate → _sysmem_new_block → malloc → brk #0` in CEF. `x23 = 0x7E9097` (1920×1080×4); `x1 = 0xA00000`, the 10 MiB chunk PartitionAlloc could not obtain.
</details>

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -56,6 +56,13 @@ GST_DEBUG_CATEGORY_STATIC (cef_console_debug);
  * width * height * 4 bytes retained per paint — 8.3 MB at 1080p — until the
  * process is killed. Two lets a new frame land while one is in flight. */
 #define MAX_QUEUED_FRAMES 2
+
+/* Same reasoning for the audio packet list, which is drained by the same
+ * create() call. Packets are far smaller than frames, so this is set well
+ * above anything normal operation produces — the list is emptied on every
+ * video frame, so it sits at a handful of entries — and only trips when the
+ * consumer has stopped entirely. */
+#define MAX_QUEUED_AUDIO_BUFFERS 128
 #define DEFAULT_URL "https://www.google.com"
 #define DEFAULT_GPU FALSE
 #define DEFAULT_CHROMIUM_DEBUG_PORT -1
@@ -391,6 +398,15 @@ class AudioHandler : public CefAudioHandler
 
     if (!src->audio_buffers) {
       src->audio_buffers = gst_buffer_list_new();
+    }
+
+    /* As with the frame queue: nothing drains this while the pipeline is not
+     * PLAYING, so drop the oldest packets rather than grow without limit. */
+    guint n_audio = gst_buffer_list_length (src->audio_buffers);
+    if (n_audio >= MAX_QUEUED_AUDIO_BUFFERS) {
+      guint drop = n_audio - MAX_QUEUED_AUDIO_BUFFERS + 1;
+      GST_DEBUG_OBJECT (src, "audio queue full, dropping %u stale buffers", drop);
+      gst_buffer_list_remove (src->audio_buffers, 0, drop);
     }
 
     gst_buffer_list_add (src->audio_buffers, buf);
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -45,6 +45,17 @@ GST_DEBUG_CATEGORY_STATIC (cef_console_debug);
 #define DEFAULT_HEIGHT 1080
 #define DEFAULT_FPS_N 30
 #define DEFAULT_FPS_D 1
+
+/* Upper bound on frames held between the browser thread and the streaming
+ * thread. CEF paints at the negotiated framerate (SetWindowlessFrameRate) and
+ * gst_cef_src_create() pops one frame per call, so in steady state the queue
+ * holds a single frame — which is the latency this element reports. The queue
+ * only grows when nothing is popping, and create() is gated behind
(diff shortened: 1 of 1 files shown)


wagenet:
This is a straight-forward enough change, but I'm not 100% confident it is the correct thing to do. It does, however, fix the issue I was encountering.

MathieuDuponchelle:
I believe https://github.com/centricular/gstcefsrc/pull/112 is more appropriate to fix your immediate issue (leak when keeping element in PAUSED), can you give it a spin and report please?

wagenet:
I had Claude do testing of this locally. I can definitely confirm that there were some crashes involved as they triggered macOS crash reports. I hate giving people raw AI outputs, but I've attempted to at least get this as clear and concise as possible. I've read through it all and it makes sense to me. But I am also very new to this codebase.

> #112 has been tested on macOS/arm64 against CEF 144. Full results are posted on #112; the summary follows.
> 
> The case it targets does work. With the pipeline parked in PAUSED, OnPaint returns before allocating and the frame queue stays empty for the whole hold — 204 frames discarded over 30 seconds, with RSS flat across the interval. For that scenario it is a better fix than a cap, since not allocating is strictly cheaper than allocating and then dropping, and it retains nothing at all rather than a couple of frames.
> 
> Three problems came out of testing, however.
> 
> The first is a crash. A pipeline that starts in PAUSED on a page producing audio aborts with SIGTRAP after roughly six seconds, reproducibly. The guard was placed in OnAudioStreamStarted rather than OnAudioStreamPacket, and because that callback returns early it never assigns mRate/mChannels, which are uninitialized members. The following OnAudioStreamPacket then sizes its allocation from that garbage, and CEF's PartitionAlloc turns the bad size into a CHECK abort rather than a NULL return. The backtrace is on #112. Neither master nor master+#111 crashes under the same test, and #111 does not mask it, since the uninitialized read has nothing to do with queue bounds.
> 
> The second is that the audio list is still unbounded in PAUSED. Nothing guards OnAudioStreamPacket, so the list grew monotonically from 1 to 321 entries across a 30-second hold and drained only on resume. This follows directly from the placement problem above — fixing the guard's location addresses both.
> 
> The third is that the frame queue still grows outside PAUSED. With #112 applied, the queue reached 34 buffers during startup while the pipeline was PLAYING, before PAUSED was ever entered. This is the structural limit of the approach: the queue grows whenever gst_cef_src_create() is not draining it, and a pipeline held in PAUSED is the most common reason for that but not the only one — blocked downstream, a stalled encoder, or a blocking pad probe produce the same condition.
> 
> On the question of which PR to take: the two look complementary rather than alternative. #112 is the right fast path for the common stall, and a cap is the backstop that bounds the queues independent of state — the two cover different failure modes and neither subsumes the other. #111 can be rebased into that narrower backstop role on top of #112, or closed if the smaller surface is preferred and #112's audio path is fixed instead. Either works; the call is yours.
> 
> One caveat on the environment: #110 had to be layered in for any of this to compile on macOS against CEF 144. Those changes are confined to CMake and GST_DEBUG format specifiers, none of them in the audio or queueing paths, and the master+#110 build without #112 does not crash — so the crash is not an artifact of them.

MathieuDuponchelle:
heh, lots of words to point out the audio check is in the wrong place :D

wagenet:
@MathieuDuponchelle true!

PR #112, opened 2026-08-11
cefsrc: discard data in the PAUSED state

MathieuDuponchelle:
cefsrc is a live element, and shouldn't queue any data until it has reached the PLAYING state.

See https://github.com/centricular/gstcefsrc/pull/111

### DIFF
--- a/gstcefsrc.cc
+++ b/gstcefsrc.cc
@@ -257,6 +257,11 @@ class RenderHandler : public CefRenderHandler
 
       GST_LOG_OBJECT (src, "painting, width / height: %d %d", w, h);
 
+      if (GST_STATE (src) == GST_STATE_PAUSED) {
+        GST_LOG_OBJECT (src, "Discarding frame in the paused state");
+        return;
+      }
+
       new_buffer = gst_buffer_new_allocate (NULL, src->width * src->height * 4, NULL);
       gst_buffer_fill (new_buffer, 0, buffer, w * h * 4);
 
@@ -320,6 +325,11 @@ class AudioHandler : public CefAudioHandler
     GstBuffer *buf;
     GstMapInfo info;
     gint i, j;
+
+    if (GST_STATE (src) == GST_STATE_PAUSED) {
+      GST_LOG_OBJECT (src, "Discarding audio packet in the paused state");
+      return;
+    }
 
     GST_LOG_OBJECT (src, "Handling audio stream packet with %d frames", frames);
 


wagenet:
_The following is a detailed Claude analysis. I always feel bad sharing raw analysis but, unfortunately, this area is a bit beyond me and it was Claude who executed all the tests. I have read through it all and it seems accurate to me as far as I can tell._

---

Test results for this PR on macOS/arm64 with CEF 144. Setup details are at the end.

The change works for the case it targets: with the pipeline parked in PAUSED, `OnPaint` returns before allocating, and the frame queue stays empty for the duration of the hold. Three problems came out of testing, one of which is a hard crash.

## 1. The audio guard is in the wrong callback, and it crashes

The guard at L306 sits in `OnAudioStreamStarted`, which appends only a small `GstEvent` to `src->audio_events`. The callback that actually accumulates data is `OnAudioStreamPacket` at L325, and it is unguarded. The log string at L307 reads "Discarding audio packet in the paused state", which suggests the hunk was meant for the latter.

The consequence is worse than the audio list simply going unbounded. `OnAudioStreamStarted` is also where `mRate` and `mChannels` are assigned, at L317-318, and the early return skips them. Both are plain `gint` members that the constructor leaves untouched — it initializes only `src` (L293, L409-410). Any `OnAudioStreamPacket` that follows therefore runs on uninitialized memory:

```c
buf = gst_buffer_new_allocate (NULL, mChannels * frames * 4, NULL);   // L336
for (i = 0; i < mChannels; i++) { gfloat *cdata = (gfloat *) data[i]; // L339 — also reads past data[]
```

A pipeline that starts in PAUSED on a page producing audio aborts as a result. The two relevant log lines appear consecutively:

```
gstcefsrc.cc:307:OnAudioStreamStarted: Discarding audio packet in the paused state
gstcefsrc.cc:334:OnAudioStreamPacket:  Handling audio stream packet with 1024 frames
<SIGTRAP>
```

The backtrace confirms the allocation as the faulting call:

```
frame #11: libgstcef.dylib`AudioHandler::OnAudioStreamPacket(...) + 100
frame #10: libgstreamer-1.0.0.dylib`gst_buffer_new_allocate + 44
frame #9:  libgstreamer-1.0.0.dylib`_sysmem_new_block + 164
frame #8..0: Chromium Embedded Framework   → EXC_BREAKPOINT
```

Because CEF installs PartitionAlloc process-wide, the invalid size becomes a `CHECK` abort rather than a NULL return, so this is not recoverable. `gst_util_uint64_scale (frames, GST_SECOND, mRate)` at L382 carries the same hazard in the event the allocation happens to succeed.

The trigger is simply a pipeline that starts in PAUSED while the page produces audio; it reproduced in all four attempts, aborting after about six seconds. Building the same test against neighbouring revisions isolates it to this PR:

| build | start-in-PAUSED, page with audio | audio packets handled |
|---|---|---|
| master | exit 0 | 321 |
| master + #111 | exit 0 | 339 |
| master + **#112** | **SIGTRAP** | 1 |
| master + #111 + #112 | **SIGTRAP** | 1 |

#111 does not mask the crash, which is expected — the uninitialized read is unrelated to whether the queues are bounded.

It is worth noting that the missing initialization of `mRate`/`mChannels` is a latent defect on master. This PR does not introduce it, but it does introduce the first code path that reaches `OnAudioStreamPacket` without having passed through the assignment, which is what makes it reachable.

## 2. Discarding the stream-start event also loses the audio caps

This is independent of the crash and would survive fixing it. The `cef-audio-stream-start` event is not merely another queued item: `cefdemux` collects it at gstcefdemux.cc:242, and it is the only input to `gst_cef_demux_update_audio_caps` at gstcefdemux.cc:196-199. If it is discarded while the element is in PAUSED, nothing re-emits it on the transition to PLAYING, so the audio pad's caps are never set for the remainder of the run.

Moving the guard to `OnAudioStreamPacket` and leaving `OnAudioStreamStarted` unguarded addresses both this and §1. Initializing `mRate`/`mChannels` in the constructor is worthwhile regardless, as defence against the latent defect noted above.

## 3. The queues still grow outside the PAUSED state

The measurement below warms the pipeline up in PLAYING, holds it in PAUSED for 30 seconds, then returns it to PLAYING, running 1080p against an animating page with looping audio.

| build | max frame queue | max audio queue | frames discarded | peak RSS |
|---|---|---|---|---|
| master | 250 | 340 | 0 | 343 MB |
| **#112** | **34** | **321** | 204 | 236 MB |
| #111 | 2 | 128 | 0 | 191 MB |
| #111 + #112 | 2 | 128 | 198 | 166 MB |

Two things stand out. The first is that audio in PAUSED is entirely unaffected: there were no discards at all across the hold, and the queue climbed monotonically from 1 to 321 entries at roughly one per 93ms, draining only once the pipeline resumed. That is the direct consequence of the placement problem in §1.

The second is that the frame queue reached 34 buffers while the pipeline was PLAYING, during startup, before PAUSED had been entered at all — around 280 MB nominal at 1080p. RSS understates this, as the synthetic test frames compress well, so queue depth is the more reliable metric of the two. This is the structural limit of a state check: the queue grows whenever `gst_cef_src_create()` is not draining it, and while a pipeline held in PAUSED is much the most common cause, blocked downstream, a stalled encoder, and a blocking pad probe all produce the same condition without ever leaving PLAYING.

## 4. Unsynchronized state read

A minor point: `GST_STATE (src)` is read on the CEF browser thread without `GST_OBJECT_LOCK`, concurrently with the thread that writes it during state changes. No misbehaviour was observed on the tested platform, and in practice this is likely benign, but it is the kind of thing TSan would report.

## Assessment

For the PAUSED case this approach is preferable to a cap. Not allocating is strictly cheaper than allocating and then dropping, and it retains nothing rather than a couple of frames, which the measurements bear out — 204 frames discarded with RSS flat across the hold.

The limitation is that it is a state check rather than a bound, so it does not address §2 or §3. The two PRs therefore look complementary rather than alternative: this one as the fast path for the common stall, and a cap as the backstop that bounds the queues independent of state. #111 could be rebased into that narrower role on top of this PR, or dropped in favour of fixing the audio path here, depending on how much surface is wanted.

§1 is the blocking issue for merging as submitted.

---

<details>
<summary>Test setup</summary>

macOS 15 / arm64, CEF 144.0.21, GStreamer 1.28.6, Apple clang 21.

All four builds also carry #110, which is required to compile on macOS against CEF 144 at all. Those changes are limited to CMake and `GST_DEBUG` format specifiers, with nothing in the audio or queueing paths, and the master + #110 build without #112 does not crash — which controls for them. Line numbers cited above are against this PR's head as submitted.

The pipeline under test was `cefsrc ! cefdemux name=d  d.video ! fakesink sync=false  d.audio ! fakesink sync=false`, pointed at a local page with an animating 1920x1080 canvas and looping audio. Queue depths were read from `cefsrc:6` logs and RSS sampled at 1Hz via `mach_task_basic_info`.

One note for anyone reproducing this on macOS: cefsrc deliberately does not call `CefRunMessageLoop()` on this platform (gstcefsrc.cc:918), relying on the host application's Cocoa run loop instead. A test harness consequently has to pump `CFRunLoopRunInMode` rather than sleep between samples — otherwise the browser never paints, and the test measures nothing while still appearing to pass.

</details>

MathieuDuponchelle:
Fixed location of the audio check. Regarding queues growing in PLAYING, this can be solved externally without adding new properties, just place leaky queues downstream.

PR #113, opened 2026-09-13
cefsrc: Add unpremultiply property

wagenet:
If I understand correctly, Chromium is rendering a premultiplied BGRA which is passed on directly by cefsrc. However, gstreamer expects conventional alpha. This change seeks to convert what Chromium provides to the more conventional format.

---

*Claude's wordy analysis*

Chromium's off-screen rendering paints premultiplied BGRA, and `cefsrc` copies `OnPaint` into buffers unchanged. Raw video in GStreamer is conventionally straight alpha, and caps cannot say otherwise, so `compositor` and `glvideomixer` blend a `cefsrc` stream too dark wherever alpha is partial: 50% white over grey 64 comes out 96 instead of 160.

A new boolean property, `unpremultiply`, converts each video buffer to straight alpha. It defaults to off, so existing output is byte-for-byte unchanged for anyone who already compensates downstream or wants premultiplied frames. I chose a boolean over an `alpha-mode` enum because the enum would have only two values, one of them the status quo. The name says what the element does, in the same style as the existing properties. It can be changed in any state and is read once per buffer under the object lock.

The conversion runs in `create()` on the streaming thread, after `queue_lock` is released. `OnPaint` runs on the CEF UI thread (the main thread on macOS), so converting there would delay Chromium's next paint. The buffer allocated by `OnPaint` has no other reference, so it is converted in place without allocating. The formula is `c' = min(255, (c*255 + a/2) / a)`, read from a 256×256 table built once in `class_init`. Pixels with alpha 0 or 255 are skipped because both encodings store them the same way.

On an M4 Max at 1080p, the loop takes 0.56 ms when all alpha is 0 and 1.23 ms when all alpha is partial. Random noise with 10% partial alpha, the worst case for branch prediction, takes 1.69 ms.

### Verification

This repo has no test harness, so verification was manual. Setup: macOS arm64, GStreamer 1.28.6, CEF 144.0.21. The pipeline was `cefsrc ! video/x-raw,format=BGRA,width=320,height=180 ! multifilesink`, run from `gst-launch-1.0` inside an app bundle. The test page has columns of each colour below over a transparent body, plus a 1px element animated with `requestAnimationFrame` so Chromium keeps painting. RGBA values were read from the last raw frame:

| CSS | master | branch, default | `unpremultiply=true` |
|---|---|---|---|
| `rgba(255,255,255,.5)` | 128,128,128,128 | 128,128,128,128 | 255,255,255,128 |
| `rgb(255,0,0)` | 255,0,0,255 | 255,0,0,255 | 255,0,0,255 |
| `rgba(255,0,0,.5)` | 128,0,0,128 | 128,0,0,128 | 255,0,0,128 |
| `rgba(255,255,255,.1)` | 26,26,26,26 | 26,26,26,26 | 255,255,255,26 |
| `rgba(255,0,0,.1)` | 26,0,0,26 | 26,0,0,26 | 255,0,0,26 |
| `rgba(200,100,50,.25)` | 50,25,13,64 | 50,25,13,64 | 199,100,52,64 |
| `rgba(10,20,30,.03)` | 0,1,1,8 | 0,1,1,8 | 0,32,32,8 |
| `transparent` | 0,0,0,0 | 0,0,0,0 | 0,0,0,0 |

At `.03`, Chromium has already quantized the colour to 0,1,1 before `cefsrc` sees it, so the original colour cannot be recovered at the source.

- **Default output unchanged:** over the whole frame, excluding the animated pixel, the default output is byte-identical to a `master` build.
- **Conversion matches the formula:** with `unpremultiply=true`, every pixel equals the formula applied to the default-mode pixel (0 mismatches out of 57,599).
- **Through `cefdemux`:** the `cefsrc ! cefdemux` path gives the same values.
- **Live toggle:** setting the property from a C host after 2 s in PLAYING switches the output at frame 59.
- **Not tested:** Linux and Windows builds.

This and #111 both edit `gstcefsrc.cc`, but a test merge applies cleanly. #111 drops stale frames in `OnPaint`, before `create()`, so dropped frames are never converted.


MathieuDuponchelle:
Hey, thanks for the proposed patch, I went looking for a bit but couldn't find definitive confirmation about whether CEF outputs straight or premultiplied alpha, do you have a source for that by any chance?

wagenet:
@MathieuDuponchelle thanks for taking a look at this. AFAICT, the docs don't specify. However, in my own observations I saw that `CSS rgba(255,255,255,.5)` arrives as `128,128,128,128`, not `255,255,255,128` which is showing premultiplied behavior. This is also reflected in the table in the PR description. This line in the ref code also refers to the values as being premultiplied: https://github.com/chromiumembedded/cef/blob/12f1636e801da935a59650e7f544fcb7ff8b8e8b/tests/cefclient/browser/osr_renderer.cc#L487-L488

